PR #13, opened 2020-04-16
Simplify AuthContextModule type

zephraph:
Simplifies the `AuthContextModule` type to ensure folks don't need to remember to manually copy things over. 

<details>
  <summary>📦 Published PR as canary version: <code>0.0.4-canary.13.179.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/cohesion@0.0.4-canary.13.179.0
  # or 
  yarn add @artsy/cohesion@0.0.4-canary.13.179.0
  ```
</details>

### DIFF
--- a/src/Schema/Authentication.ts
+++ b/src/Schema/Authentication.ts
@@ -6,51 +6,7 @@ import { ContextModule } from "./ContextModule"
 /*
  * the component where an auth modal was triggered
  */
-export type AuthContextModule =
-  | ContextModule.aboutTheWork
-  | ContextModule.artistHeader
-  | ContextModule.artistRecentlySold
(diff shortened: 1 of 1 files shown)


eessex:
The distinction between `ContextModule` and `AuthContextModule` is intentional -- we will soon be adding events outside of auth that will add a huge amount of values to `ContextModule`, but want to limit what is available to authentication modals.

zephraph:
This makes my brain itch, ha. It feels like this could be a source of bugs where one adds it correctly to the `ContextModule` but not to `AuthContextModule`. Perhaps is a non-issue though. I'll close, we can always revisit if it becomes a problem.

PR #21, opened 2020-04-22
Auth events use triggerSeconds

eessex:
Auth events use triggerSeconds rather than trigger for args.


artsyit:
:rocket: PR was released in [`v0.2.3`](https://github.com/artsy/cohesion/releases/tag/v0.2.3) :rocket:

PR #24, opened 2020-04-23
Create iOS tapped group events and schema 

mikehrom:
Wow, my first cohesion PR! Exciting! 

This new event is for the recently overhauled iOS app, especially (but not limited to) the rails and cards on the home screen. The designs I'm referring to are [here](https://www.figma.com/file/gzAi9BJi62AaJuhMRsgXMj/Home?node-id=2083%3A1974) and [here](https://www.figma.com/file/SrzeRLLcngQ78Z1Csb0BVb/Work-in-Progress-Components?node-id=1%3A2200).

I have added some new `context_module`s ('auctionCard', 'fairCard' and more). We can discuss naming but I am trying to be 1) as literal/explicit as possible and 2) as adherent to the design as possible. This PR can house the discussion for which modules we should keep/discard/add.

I've also created a schema doc for `ScreenOwnerType` which should probably be 1-1 with `util.eigen_pagetypes` but for now I've only included critical screen types.

@abhitip tomorrow, we can pair on writing the requisite test. I did not attempt that yet. Looking forward to chatting! 🎊


src/Schema/Event.ts
```diff
@@ -93,3 +105,37 @@ export interface SuccessfullyLoggedIn {
...
+ *    module_height: "single"
+ *    type: "thumbnail"
+ *  }
+ * ```
+ */
+export interface tappedEntityGroup {
+  action: ActionType.tappedEntityGroup
+  context_module: ContextModule
+  context_screen_owner_type: ScreenOwnerType
+  context_screen_owner_id?: string
```
mikehrom, line 133: does `?:` mean optional?

src/Schema/ContextModule.ts
```diff
@@ -12,17 +13,21 @@ export enum ContextModule {
   artworkImage = "artworkImage",
   artworkSidebar = "artworkSidebar",
   auctionSidebar = "auctionSidebar",
+  auctionCail = "auctionCard",
```
samchieng, line 16: @mikehrom is this supposed to be "auctionCard"?

mikehrom: Yes fixed - thank you 👍

src/Schema/Event.ts
eessex: Yes, that's right. Find more info on interfaces here: https://www.typescriptlang.org/docs/handbook/interfaces.html

src/Schema/ContextModule.ts
```diff
@@ -31,20 +36,28 @@ export enum ContextModule {
...
   mainCarousel = "mainCarousel",
   minimalCTABanner = "minimalCTABanner",
+  newWorksByArtistsYouFollowRail = "newWorksByArtistsYouFollowRail",
   otherWorksByArtistRail = "otherWorksByArtistRail",
   otherWorksFromPartnerRail = "otherWorksFromPartnerRail",
   otherWorksFromShowRail = "otherWorksFromShowRail",
   otherWorksInAuctionRail = "otherWorksInAuctionRail",
   partnerHeader = "partnerHeader",
   pastFairs = "pastFairs",
+  popularArtistsRail = "popularArtistsRail",
```
eessex, line 46: We also have `worksByPopularArtistsRail` on #61, is it possible these are duplicates?

src/Schema/ScreenOwnerType.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * All valid screen (page) types in Eigen
```
eessex, line 2: I wonder if these are specific to Eigen, do we have an equivalent in force around page owner?

src/Schema/ContextModule.ts
mikehrom: `popularArtistsRail` is one i added in reference to this rail on Eigen's home screen: 
popularartists

and i would expect this to be different than a rail of artworks that are by popular artists, hence the separation. I can for sure add descriptive comments throughout!

src/Schema/ScreenOwnerType.ts
mikehrom: Some of them are specific to Eigen (city guide screens, for example), but there will be a lot of overlap. I would imagine if we implement something similar to this event for Force (i.e. `clickedEntityGroup`), we could use the same list. The main difficulty here is a naming one: we have pages and clicks for Force and screens and taps for Eigen.

src/Schema/ContextModule.ts
mikehrom: I'm actually now wondering if `popularArtistsCardRail` is the better term, since this is a rail of cards (per the design).

src/Schema/ScreenOwnerType.ts
eessex: We can still share the owner type, and also create a second layer to limit by different contexts. Sharing this field, if there is significant overlap, will not overlap with the clicks/taps. We do something similar with `AuthContextModule`, which exposes a limited set of `ContextModule` values for use in authentication components.

src/Schema/ContextModule.ts
mikehrom: @samchieng is there a situation in which an auction/fair/artist card will standalone and not sit within a rail?

src/Schema/ScreenOwnerType.ts
abhitip: @eessex something that might be helpful for both Eigen and Force is requiring these three properties on all events: `context_screen_owner_type`, `context_screen_owner_slug`, `context_screen_owner_id`. We've discussed this before but I don't think we ever came to a final decision!

These properties will tell us what specific page (like artwork or artist) an event occurred on. We already get a `path` and `url` from Segment's automated [common fields](https://segment.com/docs/connections/spec/common/), but this extra data will help us do more focused analysis. Lmk if you think this could be automated for all events!

mikehrom: @abhitip the only reason i left `slug` and `id` optional was because of the 'Home' screen, which doesn't have either. I can see `type` being required all the time, but there might be general screens that don't have "roots."

src/Schema/ContextModule.ts
abhitip: This rail on iOS is actually the same as the "Most Popular" tab in `artistsToFollowRail` context_module on the home page on web. I also see a Trending artists rail in iOS that is the same as the "Trending" tab in the `artistsToFollowRail` on web.

Screen Shot 2020-04-23 at 2 28 20 PM

This brings up an interesting point around parity between web and iOS. What are our thoughts on removing the `artistsToFollowRail` context_module in favor of the `trendingArtistsRail` and `popularArtistsRail` context_modules that Mike added? @eessex this would require us to update the context_module for a couple of `authImpression` events.

And then @mikehrom it might be helpful for us to compare the iOS rails to the logged-in (and logged-out) home page experiences to make sure our wording makes sense so that we can then use the context_modules you're adding now for web in the future. I'd be happy to pair with you today or tomorrow if you'd like!

src/Schema/ScreenOwnerType.ts
abhitip: Yup makes sense! I have a feeling we can still automate this for all events - either we always pass all 3 properties but allow for 2 of them to be NULL or we have some if/else statement to handle it. I'll discuss it with eve!

src/Schema/ContextModule.ts
eessex: I'm open to sharing these contexts - feel free to add to the schema for `AuthContextModule` and I can remove the old ones in a separate PR.

src/Schema/ContextModule.ts
```diff
@@ -12,39 +13,53 @@ export enum ContextModule {
...
+  collectionRail = "collectionRail",
+  collectionCard = "collectionCard",
   consignSubmissionFlow = "consignSubmissionFlow",
   currentShowsRail = "currentShowsRail",
   fairInfo = "fairInfo",
   fairsHeader = "fairsHeader",
+  fairRail = "fairRail",
+  fairCard = "fairCard",
   featuredArtistsRail = "featuredArtistsRail",
+  featuredCollection = "featuredCollection",
```
mikehrom, line 33: going with this name for the following module: 
Screen Shot 2020-04-23 at 2 58 00 PM

but open to other names. the [designs](https://www.figma.com/file/gzAi9BJi62AaJuhMRsgXMj/Home?node-id=2950%3A0) refer to 'promotions' and 'promo space' but "featuredCollection" covers most cases I can think of. perhaps "featuredCollectionCard" or "featuredCollectionRail" to match what we do for featured Artists?

src/Schema/Event.ts
```diff
@@ -93,3 +105,38 @@ export interface SuccessfullyLoggedIn {
...
+  context_screen_owner_slug?: string
+  destination_screen_owner_type: ScreenOwnerType
+  destination_screen_owner_id?: string
+  destination_screen_owner_slug?: string
+  freshness?: number
```
mikehrom, line 138: pitching `freshness` to describe how old the works tapped are on the [new works-for-you screen](https://www.figma.com/file/gzAi9BJi62AaJuhMRsgXMj/Home?node-id=196%3A817): 
Screen Shot 2020-04-23 at 3 03 06 PM

For example, if you tap today's works, your `freshness` = 0, but if you tap yesterday's, it's 1. The number should be the count of days between the current_date and when the works were added to the user's feed.

src/Schema/Event.ts
```diff
@@ -93,3 +105,38 @@ export interface SuccessfullyLoggedIn {
...
+  horizontal_slide_position?: number
```
mikehrom, line 139: for example: 
Screen Shot 2020-04-23 at 3 06 55 PM

mikehrom:
I think @anipetrov may comment this but we discussed on slack briefly going down one more level of granularity on this event. So rather than `Entity` we have unique events for `tappedArtworkGroup`, `tappedArtistGroup` etc. - thoughts? I like this as well, and it sounds like then we could do a little more limiting of which specific context modules apply to each unique `Group`.

src/Schema/ContextModule.ts
samchieng: > @samchieng is there a situation in which an auction/fair/artist card will standalone and not sit within a rail?

It's something we're potentially thinking of having in the future, but nothing spec'd out yet.

src/Schema/Event.ts
samchieng: We're holding on this screen b/c I think the freshness we were trying to have here requires a heavier lift on our back-end. CC @ashfurrow

mikehrom: gotcha. we can keep out of the event for now then!

src/Schema/ScreenOwnerType.ts
```diff
@@ -0,0 +1,21 @@
...
+  CityGuideGuide = "CityGuideGuide",
+  CityGuideMap = "CityGuideMap",
+  CityPicker = "CityPicker",
+  Fair = "Fair",
+  Inbox = "Inbox",
+  Sale = "Sale",
+  Search = "Search",
+  SavesAndFollows = "SavesAndFollows",
+  ViewingRoom = "ViewingRoom",
+  ViewingRoomArtworks = "ViewingRoomArtworks",
```
mikehrom, line 19: @louislecluse i've added these screen types here re: viewing rooms, in case you need to edit or reference

src/Schema/Event.ts
```diff
@@ -126,17 +160,117 @@ export interface SuccessfullyLoggedIn {
...
   context_screen_owner_id?: string
   context_screen_owner_slug?: string
   destination_screen_owner_type: ScreenOwnerType
   destination_screen_owner_id?: string
   destination_screen_owner_slug?: string
-  freshness?: number
   horizontal_slide_position?: number
   module_height?: "single" | "double"
-  type: "button" | "thumbnail" | "header" | "stub"
+  type: "thumbnail" | "header" | "stub"
```
abhitip, line 275: What's a stub?

mikehrom: good question! it's the info underneath the thumbnail on an artist/auction/collection card, for example: 
Screen Shot 2020-04-24 at 10 27 53 AM

We could call it something different, but I think 'stub' has been used in the past. @samchieng is there an official name for it?

samchieng: I've been calling it `Card Info` but I think Stub makes sense too if we've used that before in other places

mikehrom:
Added `tappedExploreGroup` for the new City Guide entry experience. Here's its new location on the upcoming search (explore) tab: 
Screen Shot 2020-04-24 at 1 49 59 PM

src/Schema/Event.ts
```diff
@@ -93,3 +105,203 @@ export interface SuccessfullyLoggedIn {
...
+ *    action: "tappedExploreGroup"
+ *    context_module: "cityGuideCard"
+ *    context_screen_owner_type: Search
+ *    destination_screen_owner_type: CityGuideGuide
+ *    module_height: "double"
+ *    type: "thumbnail"
+ *  }
+ * ```
+ */
+export interface tappedExploreGroup {
```
mikehrom, line 261: right now, this is intended to track the new city guide entry path, but I was imagining we might have other 'explore' experiences in the future, so I left this event as open and generic as possible. 

One thing I was wondering -- right now we call the screen this card sits on the `Search` screen, but the new designs have the bottom nav referring to the page icon as `Explore`. Should we standardize here? I think `Search` is good for the screen type because it's an existing screen/page type, even if the copy/UI says otherwise.

samchieng: Ooh good call, actually we might want to refer to this page (bottom of stack) as `Explore` and the page when you tap into the search input bar `Search` (new page on top).

We'll eventually move the recent searches to the `Search` screen, plus potentially like trending searches, etc.
image

But we'll keep `explore` for just general browsing:
image

src/Schema/Event.ts
```diff
@@ -93,3 +105,203 @@ export interface SuccessfullyLoggedIn {
...
+ *    destination_screen_owner_type: Artist
+ *    destination_screen_owner_id: "5359794d1a1e86c3740001f7"
+ *    destination_screen_owner_slug: "anthony-hunter"
+ *    horizontal_slide_position: 1
+ *    module_height: "double"
+ *    type: "stub"
+ *  }
+ * ```
+ */
+export interface tappedArtistGroup {
```
damassi, line 129: This is minor but interfaces here and elsewhere are [generally CapsCase](https://github.com/artsy/cohesion/blob/master/src/Events/Authentication/CreatedAccount.ts#L10) so that they can be distinguished from common variables / functions when later imported into code.

src/Schema/ScreenOwnerType.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * All valid screen (page) types in Eigen
+ * @packageDocumentation
+ */
+export enum ScreenOwnerType {
+  Artist = "Artist",
```
eessex, line 6: We are using `lowerCamelCase` for all naming conventions except for enum/interface names.

src/Schema/Event.ts
```diff
@@ -93,3 +105,203 @@ export interface SuccessfullyLoggedIn {
...
+  action: ActionType.tappedExploreGroup
+  context_module: ContextModule
+  context_screen_owner_type: ScreenOwnerType
+  context_screen_owner_id?: string
+  context_screen_owner_slug?: string
+  destination_screen_owner_type: ScreenOwnerType
+  destination_screen_owner_id?: string
+  destination_screen_owner_slug?: string
+  horizontal_slide_position?: number
+  module_height?: "single" | "double"
```
eessex, line 271: I wonder why `module_height` and `type` are optional properties on some but not all of these interfaces?

mikehrom: inadvertent and fixed! they should be optional everywhere. the only required properties should be `context_module`, `context_screen_owner_type`, and `destination_screen_owner_type`

abhitip:
LGTM!

artsyit:
:rocket: PR was released in [`v0.2.9`](https://github.com/artsy/cohesion/releases/tag/v0.2.9) :rocket:

src/Schema/Event.ts
abhitip: Sorry I never followed up on this but tapping on the stub vs. the thumbnail take you to the same place right? Is it important to us to know which one visitors tapped on?

mikehrom: Each `type` can direct the user to a different destination within the same rail. Examples:

On the auctions rail, the header goes to the auctions home page, but thumbnails and stubs go to individual sale pages.

On the artist rails, headers aren't tappable but the thumbnails and stubs go to the artist page.

On artwork rails (specifically the 'based on other artist' rails), the header goes to an artist page, there are no stubs (presently), and the thumbnails go to artwork pages.

In short, I think it's important to know which was tapped.

abhitip: I think header is definitely important but it just seems like thumbnail vs. stub are the same thing. The reason I'm asking is because it's confusing to me which one to use for the collections rail! Will I need to specify two events for the collections rail? One for the thumbnail and one for the stub?

mikehrom: i see what you're saying - in this case, i think it's ok to use `thumbnail` for everything. there are designs where the thumbnail and the stub are separately tappable, but in the event they are not, i would use `thumbnail` to describe the tap.

abhitip: Perfect, got it!

PR #32, opened 2020-04-30
Tapped entity event helper

eessex:
Adds helper `tappedEntityGroup` to construct schema compliant events of types:
-  `tappedArtistGroup`
- `tappedArtworkGroup`
- `tappedAuctionGroup`
- `tappedCollectionGroup` 
- `tappedExploreGroup`
- `tappedFairGroup`

Also adds two missing `OwnerType` values, `collection`, and `home`. 
https://artsyproduct.atlassian.net/browse/PLATFORM-2352


cavvia:
Looks good to me – one thing I'm not sure about is the 'group' distinction here... we currently model a single `Tap` event that can take an `owner_id` and `owner_type` to represent the entity in question. Will that model remain under the new schema? Or is it to be replaced by `TappedEntity` and `TappedEntityGroup`? Maybe @abhitip can answer since @eessex is out this week.

abhitip:
@cavvia We discussed this as a team and decided that we wanted to move away from the `tap` event since it currently houses many different event types (due to the tap/success/failure schema design). The `tap` table is quite big now so we are hoping that pulling out these specific events will allow for smaller tables that are easier to query. 

And then I believe Eve built one helper for all events to make it easier for engineers, since they all have the same properties!

eessex:
@cavvia, yes @abhitip is correct - because the event type can be inferred from the destination type, we are using one helper for the engineering side. Each event will be in a separate table on the Redshift end.

artsyit:
:rocket: PR was released in [`v1.0.1`](https://github.com/artsy/cohesion/releases/tag/v1.0.1) :rocket:

PR #39, opened 2020-05-12
Sell tab tracking in iOS

louislecluse:
_My first cohesion PR_ 🔥 

# What

cc @pepopowitz

- Tracking for the new Sell tab in iOS
- Design [here](https://www.figma.com/file/yDSHlbJVzBxvSUROD2bfRI/Sell-tab-new-file?node-id=1%3A2) ("Sell tab landing - current build" section)
- events tracked [here](https://docs.google.com/spreadsheets/d/1Jo8Ns8p0N-pEWKvsfMBGY-UuU7lTQnFnrUbSjFJO5sU/edit?usp=sharing)

# Questions on the PR

- I am not extra sure of my new `context_module` names - would love some feedback here.
- `tappedConsign` event:
1. This is just a tap of a button, but I created this new event to single out consign taps (as on Force). Welcoming any thoughts here whether I should create a broader event.
2. I added a subject property. I added this property **in the event file** and not in a separate property file such as `context_module`. Is it OK?

# Questions on Cohesion

- What is the difference between the general `tap` event and all the "sub-events" in it? (`tappedArtistGroup`, etc.). My underlying question is when should I branch out of a generic event?

# Next Steps
- Let me know if this PR works or if I am totally wrong. In the latter case I could pair with @abhitip ?
- @pepopowitz I think you might have to create a new event for `tappedConsign` in the `/Events` folder?

# Closing Remarks

This PR has been very very smooth to write big props @abhitip and @eessex !!!

# Todo

- [x] Refine context modules for tapped consign
~~- [ ] Subject properties for `tappedArtworkGroup` and `tappedArtistGroup`~~


src/Schema/Events/Tap.ts
```diff
@@ -179,3 +179,30 @@ export interface TappedEntityGroup {
...
+ *    destination_screen_owner_type: "consignmentSubmission",
+ *    subject: "startSelling"
+ *  }
+ * ```
+ */
+export interface TappedConsign {
+  action: ActionType.tappedConsign
+  context_screen_owner_type: ScreenOwnerType
+  destination_screen_owner_type: ScreenOwnerType
+  subject: Subject
```
eessex, line 202: If the subject is derived from the button text, I think it would be preferable to use an untyped `string` instead of a manually defined type. This way we can pass the button text directly from the UI to segment, which will remove steps in case a button's text is changed.

mikehrom:
@louislecluse  the `tappedConsign` event is sourced from a button _within_ the sell tab, correct? versus the tapping of the 'sell' icon within the bottom nav? Just checking, since I think the bottom nav taps could comprise their own tap event/module.

src/Schema/Event.ts
```diff
@@ -59,6 +60,11 @@ export enum ActionType {
    * Corresponds to {@link TappedFairGroup}
    */
   tappedFairGroup = "tappedFairGroup",
+
+  tappedConsign = "tappedConsign",
+  /**
```
eessex, line 65: This comment should be above the code it describes.

eessex: Also it corresponds to `TappedConsign` (uppercase) -- the schema this relates to

src/Schema/Events/Tap.ts
pepopowitz: +1 to this suggestion, especially since the designs I'm looking at only have a "start selling" button, but not a "start today" button.

pepopowitz: Also - since the buttons at the top and bottom of the designs both say "start selling", do we need some other field to identify which of the two buttons they clicked on?

louislecluse: - Good point will put a string type
- @pepopowitz: agreed. We can use a context module.
What are the modules called in the code? I am thinking for the one at the top of the page "sellHeader" but I am unsure about the one at the bottom. Is there any name in the code for that footer header?

src/Schema/Event.ts
louislecluse: Thanks!

louislecluse:
Additionally @pepopowitz I think the context modules for `tappedArtworkGroup` and `tappedArtistGroup` are too detailed ([spreadsheet](https://docs.google.com/spreadsheets/d/1Jo8Ns8p0N-pEWKvsfMBGY-UuU7lTQnFnrUbSjFJO5sU/edit?usp=sharing))

What is the name of the context modules for the artist grid and the recently sold works grid in the sell tab? We may want to use these as context modules. 
I can then add a `subject` property that describes the exact text for each grid (eg: "Artists collectors are looking to buy" for the artist grid).

src/Schema/Events/Tap.ts
pepopowitz: `sellHeader` and `sellFooter` would be good. We call them `HeaderCTA` and `FooterCTA` in the code but it really doesn't matter - anything to distinguish header from footer is great.

pepopowitz:
Given the detailed names in https://github.com/artsy/cohesion/blob/master/src/Schema/ContextModule.ts, I actually think the names you provided for context modules are great. They clearly identify the rail/grid in question. We don't have any context modules attached to the code yet - we won't until we implement the tracking.

src/Schema/Events/Tap.ts
eessex: @abhitip may have suggestions on `ContextModule`

src/Schema/Events/Tap.ts
```diff
@@ -179,3 +179,25 @@ export interface TappedEntityGroup {
 
 export type EntityModuleHeight = "single" | "double"
 export type EntityModuleType = "thumbnail" | "header" | "stub"
+
+/**
+ * A user clicks a Consign button in iOS
```
eessex, line 184: #trivial but might want to use `taps` over `clicks` here for clarity.

src/Schema/Events/Tap.ts
abhitip: `sellHeader` and `sellFooter` sound good to me!

src/Schema/Events/Tap.ts
```diff
@@ -179,3 +179,26 @@ export interface TappedEntityGroup {
...
+ * A user taps a Consign button in iOS
+ *
+ * This schema describes events sent to Segment from [[tappedConsign]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "tappedConsign",
+ *    context_screen_owner_type: "sell",
+ *    destination_screen_owner_type: "consignmentSubmission",
```
abhitip, line 193: Make sure you add `context_module` here!

abhitip:
Had one minor comment but otherwise this looks great to me! New context_modules and adding the `subject` field makes a lot of sense to me :)

abhitip:
Oh also, `action` should not be a special property that is passed to segment. It's just what engineers use to indicate the `event_name`. I noticed `action` in your spreadsheet and just wanted to make sure that you were aware of that!

abhitip:
LGTM!

artsyit:
:rocket: PR was released in [`v1.0.4`](https://github.com/artsy/cohesion/releases/tag/v1.0.4) :rocket:

eessex:
I'll follow up to add a helper 🎉

PR #48, opened 2020-05-21
Expose consign helpers

jonallured:
I might be missing something, so please forgive me if so!!

But when I went to use the `tappedConsign` and `TappedConsignArgs` in Eigen, I was getting errors about those things not existing in `@artsy/cohesion` so I think this is what's required to do that? LMK if there's a better way!


artsyit:
:rocket: PR was released in [`v1.1.0`](https://github.com/artsy/cohesion/releases/tag/v1.1.0) :rocket:

PR #50, opened 2020-05-22
Don't require module height for entity group

jonallured:
Per some conversation over on this one: https://github.com/artsy/eigen/pull/3350#discussion_r429335663 this PR marks the module height bit as optional so that I can leave it off over there.


damassi:
(Eve is OOO today)

artsyit:
:rocket: PR was released in [`v1.3.0`](https://github.com/artsy/cohesion/releases/tag/v1.3.0) :rocket:

PR #51, opened 2020-05-22
Fix broken build

jonallured:
I didn't notice that my PRs have not been building because they've come from forks so I was able to sneak through a broken master build. 😭 

I caught this only after trying to update Eigen to not provide a module height and discovered that the types were out-of-sync. So I went back to my local clone, updated to latest master and discovered the type error. Then I fixed it. Then I looked again at the Circle setup and discovered the forks are not built.

So I'm opening this from upstream and I think that will cause a build on Circle and that build should be green. If all of that is true, then this can be merged, a new version will get deployed and Eigen will pick up that new version and then I can rebase my work on top of that and be able to skip providing a module height! 😅

### DIFF
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -173,7 +173,7 @@ export interface TappedEntityGroup {
   destination_screen_owner_id?: string
   destination_screen_owner_slug?: string
   horizontal_slide_position?: number
-  module_height: EntityModuleHeight
+  module_height?: EntityModuleHeight
   type: EntityModuleType
 }
 


jonallured:
Ok I'm super confused now, this PR from upstream didn't trigger a CI run either. 🤔

artsyit:
:rocket: PR was released in [`v1.3.0`](https://github.com/artsy/cohesion/releases/tag/v1.3.0) :rocket:

jonallured:
Welp master is back to green:

https://app.circleci.com/pipelines/github/artsy/cohesion/189/workflows/e22f3ed6-dc4a-484f-bd0e-590e0b2521e3

So that's good, but there's a follow up item here to figure out what's up with PRs not building on CI. 🤷

PR #55, opened 2020-06-04
Adds ClickedEntityGroup events

abhitip:
This adds the click events for web; they pretty much have the same schema as the tap events. I also added a context module for the new `topWorksRail` on the artist pages that FX is working on. One question - I did not keep `module_height` for the click events. Do we think we need it on web?


src/Schema/Events/Click.ts
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * This schema describes events sent to Segment from [[clickedArtworkGroup]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "clickedArtworkGroup",
+ *    context_module: "newWorksByArtistsYouFollowRail",
+ *    context_screen_owner_type: "home",
+ *    destination_screen_owner_type: Artwork,
```
eessex, line 52: destination screen owner type should be `"artwork"`

src/Schema/Events/Click.ts
```diff
@@ -0,0 +1,152 @@
...
+ */
+export interface ClickedEntityGroup {
+  action:
+    | ActionType.clickedArtistGroup
+    | ActionType.clickedArtworkGroup
+    | ActionType.clickedAuctionGroup
+    | ActionType.clickedCollectionGroup
+    | ActionType.clickedFairGroup
+  context_module: ContextModule
+  context_screen_owner_type: PageOwnerType
```
eessex, line 144: just double checking that these are `context_screen_owner_type` vs `context_page_owner_type`

abhitip: oo good catch - changing now

mikehrom:
@abhitip on 'module_height' i think design is best to consult on that ("will we be interested in the visual treatment of modules and their effectiveness of converting traffic?") but i also think if we include it at all, it should be optional. not noticing a lot of early appetite to track it.

abhitip:
Okay got it! I'll leave it out for now and we can revisit later.

artsyit:
:rocket: PR was released in [`v1.4.0`](https://github.com/artsy/cohesion/releases/tag/v1.4.0) :rocket:

PR #56, opened 2020-06-05
Adds timeOnPage

abhitip:
As a part of the Segment audit, we decided to remove `timeOnPage` events for 30 seconds, 1 min, and 3 mins. We are also using this as an opportunity to move `timeOnPage` into cohesion.

The current `time_on_page` event has two properties - `category` for the time and `message` for the page path. I decided to get rid of `message` in favor of the `context_page_owner_[x]` properties. Lmk what you think!

### DIFF
--- a/src/Schema/Events/System.ts
+++ b/src/Schema/Events/System.ts
@@ -1,0 +1,29 @@
+import { ActionType } from "../Event"
+import { PageOwnerType } from "../OwnerType"
+
+/**
+ * Schemas describing system events
+ * @packageDocumentation
+ */
+
+/**
+ * A user has been on a page for 15 seconds.
+ *
+ * This schema describes events sent to Segment from [[timeOnPage]]
+ *
+ *  @example
+ *  ```
+ *  {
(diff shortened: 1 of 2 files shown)


src/Schema/Events/System.ts
```diff
@@ -0,0 +1,29 @@
...
+ *  {
+ *    action: "timeOnPage",
+ *    category: "15 seconds",
+ *    context_page_owner_type: "home"
+ *  }
+ * ```
+ */
+export interface TimeOnPage {
+  action: ActionType.timeOnPage
+  category: "15 seconds"
```
eessex, line 25: Is 15 seconds the only option available?

eessex:
Pulling the latest from master should resolve your conflicts.

src/Schema/Events/System.ts
abhitip: Yes it's the only event we want to fire for right now! We might decide to add another time option later, but for now, we are trying to cut back on events since we keep maxing out our Segment API limit.

PR #57, opened 2020-06-08
Add clickedEntityGroup helper

eessex:
Adds helper `clickedEntityGroup` to support new click events.


src/Schema/OwnerType.ts
```diff
@@ -62,5 +62,8 @@ export type PageOwnerType =
   | OwnerType.collection
   | OwnerType.consign
   | OwnerType.fair
+  | OwnerType.gene
```
eessex, line 65: @abhitip - wasn't sure what contexts this will be used in, but we may need to expand available `PageOwnerTypes` for web while implementing.

abhitip: Yes we can expand as we go!

PR #58, opened 2020-06-10
Adds clickedMainArtworkGrid

abhitip:
This adds `clickedMainArtworkGrid`. I added a comment in the code, but I decided to create another event for the artwork grid (instead of using clickedArtworkGroup) because I think it's important and frequent enough that it'll make it easier to access if it's in its own table. Lmk what you think!

### DIFF
--- a/src/Schema/ContextModule.ts
+++ b/src/Schema/ContextModule.ts
@@ -11,7 +11,7 @@ export enum ContextModule {
   artistRecentlySold = "artistRecentlySold",
   artistsTab = "artistsTab",
   artistsToFollowRail = "artistsToFollowRail",
-  artworkGrid = "artworkGrid",
+  mainArtworkGrid = "mainArtworkGrid",
   artworkImage = "artworkImage",
   artworkRecentlySoldGrid = "artworkRecentlySoldGrid",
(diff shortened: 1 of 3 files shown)


mzikherman:
Is this ready to go? (wondering b/c I'm looking to add some tracking code in Force that uses this prop 🙈 )

eessex:
@mzikherman I still need to add a helper for use in force, can update you as soon as we're ready to use!

mikehrom:
I think this _is_ ready to go, LGTM. as a note of clarification, we think the 'main grid' component (on collect, collection, artist and search results pages) is prominently used enough to warrant its own event name, as opposed to being grouped in as a context module with the `clickedArtworkGroup` event. 

at some point, a `tappedMainArtworkGrid` event will also be needed, similar to this one, but for iOS.

artsyit:
:rocket: PR was released in [`v1.7.0`](https://github.com/artsy/cohesion/releases/tag/v1.7.0) :rocket:

mzikherman:
> I still need to add a helper for use in force, can update you as soon as we're ready to use!

@eessex thanks! Feel free to tag me on that.

PR #61, opened 2020-06-15
Refactors clickedMainArtworkGrid

abhitip:
We realized that `clickedMainArtworkGrid` does not need to share a schema with `clickedEntityGroup` so it makes sense to separate it out. This will also allow us to add grid-specific properties like `ranking_algo` if we want to in the future!

I also made a couple of changes to the schema:

- `type` = "thumbnail" since it currently will never not be thumbnail
- I made the destination properties required, since it will always be an artwork page

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -39,33 +39,6 @@ export interface ClickedArtistGroup extends ClickedEntityGroup {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in [`v1.8.0`](https://github.com/artsy/cohesion/releases/tag/v1.8.0) :rocket:

PR #63, opened 2020-06-15
@eessex => [OwnerType] Add collect page

mzikherman:
Adding collect page as an owner type.

### DIFF
--- a/src/Schema/OwnerType.ts
+++ b/src/Schema/OwnerType.ts
@@ -10,6 +10,7 @@ export enum OwnerType {
   cityGuideGuide = "cityGuideGuide",
   cityGuideMap = "cityGuideMap",
   cityPicker = "cityPicker",
+  collect = "collect",
   collection = "collection",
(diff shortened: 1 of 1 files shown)


mzikherman:
I don't follow when one would use that vs `OwnerType`? For instance, I'm working on a Force PR that uses `OwnerType.artist` in tracking code, should that be `PageOwnerType` instead?

eessex:
So the `OwnerType` is what you'd use in Force etc when implementing this. However, we use `PageOwnerType` on the schema side to enforce which `OwnerTypes` are exposed in web versus iOS environments.

mzikherman:
Ah, gotcha! That makes a lot of sense.

mzikherman:
Updated!

artsyit:
:rocket: PR was released in [`v1.10.0`](https://github.com/artsy/cohesion/releases/tag/v1.10.0) :rocket:

PR #68, opened 2020-06-25
Events for conversations

anipetrov:
Adding to events for the conversations tab, `focusedOnConversationMessageInput` and `sentConversationMessage`!


src/Schema/Events/Conversations.ts
```diff
@@ -0,0 +1,48 @@
...
+ *    action: "sentConversationMessage",
+ *    impulse_conversation_id: 198,
+ *    impulse_message_id: 8765
+ *    
+ *  }
+ * ```
+ */
+export interface SentConversationMessage {
+    action: ActionType.sentConversationMessage
+    impulse_conversation_id?: number
```
eessex, line 44: Is there an instance where a conversation or message does not have an `id`?

anipetrov: it should actually always have it, so i can remove the `?`!
also, i am resolving the conflicts but getting a yarn error, anyone available for a quick pair?

anipetrov: nvm fixed it!

anipetrov: i do also wonder if `number` is the right type given that these are postgresql ids?

src/Schema/Events/Conversations.ts
```diff
@@ -0,0 +1,46 @@
...
+ *  {
+ *    action: "focusedOnConversationMessageInput",
+ *    impulse_conversation_id: 198
+ *
+ *  }
+ * ```
+ */
+export interface FocusedOnConversationMessageInput {
+  action: ActionType.focusedOnConversationMessageInput
+  impulse_conversation_id?: number
```
starsirius, line 24: @anipetrov Do we also need a new context module or it's fine not supplying that for now?

src/Schema/Events/Conversations.ts
eessex: If redshift receives them as `number` then this type makes sense. And yes, the `?` being removed will make the property required.

src/Schema/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from "./Events"
 export * from "./Events/Authentication"
+export * from "./Events/Conversations"
 export * from "./Events/Click"
```
starsirius, line 4: @eessex A organization questions for events, looks like we have both actions (e.g. Click, Tap, etc.) and functionalities (e.g. Authentication, Conversations, etc.). Curious about guidelines for organizing new events.

eessex: I think conversations makes sense if we expect more events to exist in this domain.

src/Schema/Events/Conversations.ts
anipetrov: no i dont think we need one!

starsirius:
Thanks @anipetrov and @eessex for the review!

artsyit:
:rocket: PR was released in [`v1.14.0`](https://github.com/artsy/cohesion/releases/tag/v1.14.0) :rocket:

src/Schema/Events/Conversations.ts
```diff
@@ -0,0 +1,46 @@
...
+ *    impulse_conversation_id: 198,
+ *    impulse_message_id: 8765
+ *
+ *  }
+ * ```
+ */
+export interface SentConversationMessage {
+  action: ActionType.sentConversationMessage
+  impulse_conversation_id?: number
+  impulse_message_id?: number
```
eessex, line 45: It looks like these properties are still optional, but should be required.

anipetrov: ohhh i thought i pushed a commit to fix, will follow up

PR #69, opened 2020-06-25
Track taps on the iOS promo space

mikehrom:
[cc: @ashfurrow @brainbicycle]

Submitting analytics tracking for the home screen promo space on iOS. Comments within the code.

[[Jira](https://artsyproduct.atlassian.net/browse/MX-336)]

### DIFF
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -220,9 +220,8 @@ export interface TappedConsign {
  *    context_screen_owner_type: "home",
  *    destination_screen_owner_type: "collection",
  *    destination_screen_owner_slug: "artists-impacted-museum-shows",
- *    destination_path: "/collection/artists-impacted-museum-shows"
- *    subject: "Explore the collection",
- *    utm_campaign: "art-keeps-going"
+ *    destination_path: "/collection/artists-impacted-museum-shows",
+ *    subject: "Explore the collection"
  *  }
  * ```
  */
@@ -235,5 +234,4 @@ export interface TappedPromoSpace {
   destination_screen_owner_slug?: string
   destination_path?: string
   subject: string
-  utm_campaign: string
 }
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -219,7 +219,8 @@ export interface TappedConsign {
  *    context_module: "promoSpace",
  *    context_screen_owner_type: "home",
  *    destination_screen_owner_type: "collection",
- *    destination_screen_owner_slug: artists-impacted-museum-shows,
+ *    destination_screen_owner_slug: "artists-impacted-museum-shows",
+ *    destination_path: "/collection/artists-impacted-museum-shows"
  *    subject: "Explore the collection",
  *    utm_campaign: "art-keeps-going"
  *  }
@@ -232,6 +233,7 @@ export interface TappedPromoSpace {
   destination_screen_owner_type: ScreenOwnerType
   destination_screen_owner_id?: string
   destination_screen_owner_slug?: string
+  destination_path?: string
   subject: string
   utm_campaign: string
 }
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -229,9 +229,9 @@ export interface TappedPromoSpace {
   action: ActionType.tappedPromoSpace
   context_module: ContextModule
   context_screen_owner_type: ScreenOwnerType
-  destination_screen_owner_type: ScreenOwnerType
+  destination_screen_owner_type?: ScreenOwnerType
   destination_screen_owner_id?: string
   destination_screen_owner_slug?: string
-  destination_path?: string
+  destination_path: string
   subject: string
 }
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -206,3 +206,32 @@ export interface TappedConsign {
   /** The text of the tapped button */
   subject: string
 }
+
+/**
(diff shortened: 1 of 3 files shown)


src/Schema/Events/Tap.ts
```diff
@@ -206,3 +206,32 @@ export interface TappedConsign {
...
+ *  @example
+ *  ```
+ *  {
+ *    action: "tappedPromoSpace",
+ *    context_module: "promoSpace",
+ *    context_screen_owner_type: "home",
+ *    destination_screen_owner_type: "collection",
+ *    destination_screen_owner_slug: artists-impacted-museum-shows,
+ *    subject: "Explore the collection",
+ *    utm_campaign: "art-keeps-going"
```
mikehrom, line 224: `utm_campaign` is perhaps the only property we won't need, but I included it because I'd imagine the marketing organization is going to want to keep track of engagement per campaign, and this is an easy way to do it. 

I spoke with them earlier though, and they said they are not currently 100% sure how they will operationalize the promo space yet, so I could also see a world in which we add this later or make it optional. If anyone has thoughts, let me know.

mikehrom:
Designs for the promo space can be found [here](https://www.figma.com/file/gzAi9BJi62AaJuhMRsgXMj/Home?node-id=3293%3A5496).

src/Schema/Events/Tap.ts
```diff
@@ -206,3 +206,32 @@ export interface TappedConsign {
...
+ *  }
+ * ```
+ */
+export interface TappedPromoSpace {
+  action: ActionType.tappedPromoSpace
+  context_module: ContextModule
+  context_screen_owner_type: ScreenOwnerType
+  destination_screen_owner_type: ScreenOwnerType
+  destination_screen_owner_id?: string
+  destination_screen_owner_slug?: string
```
ashfurrow, line 234: I don't think we'll have this info. The promo page can point to any URL path at Artsy, from editorial content to auctions to a bespoke page like the Art Keeps Going campaign. I might be wrong, @brainbicycle and @ds300 please chime in, but I think all we'll get is a path. 

@mikehrom would it be okay to exclude these three `destination` attributes and use something like `destination_path` instead?

mikehrom: Interesting! So it will be pointing to non-native views? Will it also be capable of pointing to native screens within the app? If it can do both, we can list properties for both and make them all optional.

ashfurrow: Yeah, my understanding is that it can point to anything. We have a "router" in the app that tries to find a native screen to show for the URL, like fairs and artists and whatnot. If it can't find a native view for a given URL, then it falls back to a web view showing that URL. This is how we sent out the Art Keeps Going campaign with no native screen: it used a web view.. We could infer the content based on the URL, but I don't think that we will know what kind of entity it is. Let me know what I can clarify here.

mikehrom: Alright so I've added `destination_path` as an additional, optional property. The only counter-argument that comes to mind for this right now is that we can more or less ascertain the `path` by combining `destination_screen_owner_type` and `destination_screen_owner_slug` -- but curious what others on this PR think about that. I think the current commit reflects the most flexible way of achieving what we want.

abhitip: It sounds like @ashfurrow is saying that we won't be able to get `destination_screen_owner_type` for non-native views? If that's the case, then @mikehrom, what do you think about making `destination_path` required and the other 3 `destination_screen_owner_` properties optional?

src/Schema/Events/Tap.ts
abhitip: Would the `utm_campaign` be available in the subsequent screen's path? Maybe we don't need a property here and we can just use the screen path like we do for other channels?

mikehrom: Good point. I've removed it! 👍

src/Schema/Events/Tap.ts
mikehrom: I think screen owner type should always be mandatory - even if we're sending people to non-native views, we should know the page/screen type. Does that mean it should point to `PageOwnerType`? 

I figured `path` would be optional in the event of the promo space sending people to native iOS views, which don't have proper url-style paths.

ashfurrow: To clarify, we _only_ have a URL/path. We could write code to infer the owner type from that path, but it would be a bit of work, and brittle.

eessex: We split the path for type and slug on the web side for page view events currently.

brainbicycle: 🤔 I suppose we could use loadPath in Switchboard to query which view a given url would land on if it is a react native or native view but then we still need the owner type for any given view, I agree it is non-trivial. For web view maybe just a generic "web view" screen owner and the url should be enough context.

ashfurrow: I like what Eve said about splitting on the URL, which might be an easy solution _as long as_ the owner type is optional. So `/campaign/art-keeps-going` would be `"campaign"`, for example. But vanity URLs like `/heritage` would kind of break it, which is why I'd like it to be optional.

mikehrom: Checking back in here -- that all makes sense. I've combined feedback and left `path` mandatory (we need _something_ from the destination to be mandatory) and left the screen owner optional, for cases like @ashfurrow  described above. 👍

artsyit:
:rocket: PR was released in [`v1.16.0`](https://github.com/artsy/cohesion/releases/tag/v1.16.0) :rocket:

PR #75, opened 2020-07-02
Add context to auto jobs

eessex:
Auto jobs are failing due to node version complaints, added context to failing jobs to match Reaction's functional setup with node 12.


icirellik:
Issue appears to be resolved

PR #78, opened 2020-07-06
Track tab bar taps on iOS

mikehrom:
- implements new tracking for taps on the icons in the tab bar

I wrote a helper and tests for this mostly by copying patterns from the other helpers/tests - so definitely check my work. I'll leave some in-line comments, too.


src/Helpers/Tap/TappedTabBar.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @example
+ * ```
+ * tappedTabBar({
+ *   badge: false,
+ *   tab: "sell"
+ * })
+ * ```
+ */
+export const tappedTabBar = ({ badge }: TappedTabBarArgs): TappedTabBar => {
```
mikehrom, line 28: if i try to add `tab` to the args list here (after `badge`),  I get an error if i supply an explicit value for `tab:` on line 34. Which is correct? The current syntax or... do I add `tab` to the args list on line 28 and change what's inside the `return` brackets?

src/Schema/Events/Tap.ts
```diff
@@ -235,3 +236,27 @@ export interface TappedPromoSpace {
...
+ *    badge: true
+ *    context_module: "tabBar",
+ *    context_screen_owner_type: "home",
+ *    tab: "messages"
+ *  }
+ * ```
+ */
+export interface TappedTabBar {
+  action: ActionType.tappedTabBar
+  badge?: boolean
```
mikehrom, line 258: up for debate whether we should have this, so I've left it optional, but my thinking is: "What if we want to know a user is tapping an icon with a badge on it?" Do we foresee this being important? Otherwise, I'm ok to remove this property throughout.

src/Schema/Values/Tab.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Master list of tabs available for the iOS app Tab Bar (footer navigation)
+ *
+ * @packageDocumentation
+ */
+
+export type Tab =
```
mikehrom, line 7: The current set of values (as of 6.4.9) should be "home", "search", "messages", "sell" and "favorites" -- from left to right, on the tab bar. This is the existing naming of each tab that have been using on the event `eigen_production.profile_view`

src/Helpers/Tap/TappedTabBar.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  ActionType,
+  ContextModule,
+  OwnerType,
+  TappedTabBar,
+} from "../../Schema"
+
+export interface TappedTabBarArgs {
+  badge: boolean
+  tab: string
```
ds300, line 10: This should use the `Tab` type you defined elsewhere

```suggestion
  tab: Tab
```

src/Helpers/Tap/TappedTabBar.ts
ds300: yep you need to add `tab` here and then change line 34
```diff
     context_screen_owner_type: OwnerType.home,
-    tab: "sell",
+    tab,
   }
```

src/Schema/Values/Tab.ts
ds300: yep this covers our bases. We probably won't use `cityGuide` or `favorites` ever again but it might be good to keep them for historical context.

src/Schema/Events/Tap.ts
eessex: Let's make this required, we can set the default to "false" in the helper so that this is always included.

src/Helpers/Tap/TappedTabBar.ts
```diff
@@ -0,0 +1,36 @@
...
+ * tappedTabBar({
+ *   badge: false,
+ *   tab: "sell"
+ * })
+ * ```
+ */
+export const tappedTabBar = ({ badge }: TappedTabBarArgs): TappedTabBar => {
+  return {
+    action: ActionType.tappedTabBar,
+    badge,
```
eessex, line 31: Use `badge: badge || false` to get a default value.

src/Helpers/Tap/TappedTabBar.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  ActionType,
+  ContextModule,
+  OwnerType,
+  TappedTabBar,
+} from "../../Schema"
+
+export interface TappedTabBarArgs {
+  badge: boolean
```
eessex, line 9: `badge?: boolean` so we won't have to pass this unless it is `true`.

mikehrom:
@ds300 @eessex thanks for that! I've incorporated changes-

src/Helpers/Tap/TappedTabBar.ts
```diff
@@ -0,0 +1,41 @@
...
+ * A user taps an icon in the Tab Bar within the iOS app
+ *
+ * badge is true if present, false if not
+ *
+ * tab describes the icon tapped by the user
+ *
+ * @example
+ * ```
+ * tappedTabBar({
+ *   badge: false,
```
eessex, line 25: Use `badge: true` here so it is clear we only need to pass when true.

eessex:
You will also need to add a global export here: https://github.com/artsy/cohesion/blob/master/src/Helpers/index.ts

artsyit:
:rocket: PR was released in [`v1.20.0`](https://github.com/artsy/cohesion/releases/tag/v1.20.0) :rocket:

PR #80, opened 2020-07-15
Makes context page properties optional in timeOnPage

abhitip:
We have to make the context page properties optional due to the same issue we had [here](https://github.com/artsy/eigen/pull/3517).

We do not need to add a `path` property because `context_page_path` is already provided as a common field from Segment.


abhitip:
No...I haven't done anything. I'm so confused lol

artsyit:
:rocket: PR was released in [`v1.21.0`](https://github.com/artsy/cohesion/releases/tag/v1.21.0) :rocket:

PR #89, opened 2020-08-05
no need for sort-keys

pvinis:
This rule causes keys to go alphabetically, which doesn't offer any benefit for us here.

It's much more useful to be able to group keys in a way that makes sense, like having the `type` right next to the `context_page`. `type` is what we click on the page we are now, so it makes sense to be close to there, rather than last in the list because it is named `type` and not like `owner_page_type_of_thing_i_clicked`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,7 +20,6 @@ module.exports = {
   root: true,
   rules: {
     "no-console": 1,
-    "sort-keys": 1,
     "sort-imports": 1,
     "spaced-comment": ["error", "always", { block: { balanced: true } }],
   },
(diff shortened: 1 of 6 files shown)


eessex:
Can you explain why it is is not a benefit? Alphabetical was an intentional decision, to make it easy to find keys— especially in large objects.  I personally find the benefit sorting by type to be often short lived (next engineer might remove for who knows why) and not scale easily when there are many keys. I would prefer to leave as is.

eessex:
Also just want to clarify that the destination is the thing you clicked, context is the page which displays the rail, so destination would be the item that has a type.

pvinis:
So when I click on an artwork item on a rail in the home screen, the context is `home`, the destination is `artwork`, and the type is `thumbnail`. I don't think the type is related to the destination in any way, right? I am clicking a `thumbnail` at `home` in order to go to the `artwork`. Am I completely misunderstanding this?

abhitip:
That's correct @pvinis! Keeping it alphabetical allows for some sort of enforced consistency across all schemas - we know how to order properties at all times. I agree with you that it would be nice if it made more logical sense, but I think it would be difficult to enforce which would lead to inconsistency, at least on the analyst side. As an analyst, it's nice to know what is expected in the schema and to not have to make many decisions.

pvinis:
Cool. I'll close this.

PR #90, opened 2020-08-06
[AS-1437] Artist series analytics tracking for iOS and web

abhitip:
This adds three new events for iOS in cohesion: tappedArtistSeriesGroup, followedArtist, and unfollowedArtist. You can find the events I'd like to track for artist series in iOS [here](https://docs.google.com/spreadsheets/d/1_VDAmRJz2K87FXDq7oh_tQSgwBv4syuZjaaR3sfetcI/edit#gid=1858087022).

Note that adding followedArtist and unfollowedArtist to cohesion means we are moving follow events out of the `success` schema.

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -1,7 +1,7 @@
 import { ActionType } from "."
 import { ContextModule } from "../Values/ContextModule"
 import { EntityModuleType } from "../Values/EntityModuleType"
-import { PageOwnerType } from "../Values/OwnerType"
+import { OwnerType, PageOwnerType } from "../Values/OwnerType"
 
 /**
  * Schemas describing Click events
@@ -39,7 +39,7 @@ export interface ClickedArtistGroup extends ClickedEntityGroup {
 }
 
 /**
- * A user clicks a grouping of artist series on iOS. If the artist series was boosted by the curation team, then featured will be set to true.
+ * A user clicks a grouping of artist series on web. If the artist series was boosted by the curation team, then curation_boost will be set to true.
  *
  *  This schema describes events sent to Segment from [[clickedEntityGroup]]
  *
@@ -54,7 +54,7 @@ export interface ClickedArtistGroup extends ClickedEntityGroup {
  *    destination_page_owner_type: "artistSeries",
  *    destination_page_owner_id: "5359794d1a1e86c3740001f7",
  *    destination_page_owner_slug: "alex-katz-black-dress",
- *    featured: true,
+ *    curation_boost: true,
  *    horizontal_slide_position: 1,
  *    type: "thumbnail"
  *  }
@@ -62,6 +62,7 @@ export interface ClickedArtistGroup extends ClickedEntityGroup {
  */
 export interface ClickedArtistSeriesGroup extends ClickedEntityGroup {
   action: ActionType.clickedArtistSeriesGroup
+  destination_page_owner_type: OwnerType.artistSeries
 }
 
 /**
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -39,6 +39,32 @@ export interface ClickedArtistGroup extends ClickedEntityGroup {
 }
 
 /**
+ * A user clicks a grouping of artist series on iOS. If the artist series was boosted by the curation team, then featured will be set to true.
+ *
+ *  This schema describes events sent to Segment from [[clickedEntityGroup]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "clickedArtistSeriesGroup",
+ *    context_module: "moreSeriesByThisArtist",
+ *    context_page_owner_type: "artistSeries",
+ *    context_page_owner_id: "5359794d1a1e86c3740001f7",
+ *    context_page_owner_slug: "alex-katz-departure",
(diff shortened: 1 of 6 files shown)


abhitip:
FYI I just added web tracking to this PR as well.

I'm going to meet with @eessex to discuss how to add pageview, screenview, followedArtist and unFollowedArtist to cohesion.

src/Schema/Events/Tap.ts
```diff
@@ -177,6 +204,7 @@ export interface TappedEntityGroup {
   destination_screen_owner_type: ScreenOwnerType
   destination_screen_owner_id?: string
   destination_screen_owner_slug?: string
+  featured?: boolean
```
eessex, line 207: Historically the key `featured` has been fraught in the codebase, we have many instances of this being used with completely different meanings, which becomes confusing over times. Is there a more specific name possible? What does a featured item in the group look like?

src/Schema/Events/Click.ts
```diff
@@ -38,6 +38,32 @@ export interface ClickedArtistGroup extends ClickedEntityGroup {
...
+ *    destination_page_owner_id: "5359794d1a1e86c3740001f7",
+ *    destination_page_owner_slug: "alex-katz-black-dress",
+ *    featured: true,
+ *    horizontal_slide_position: 1,
+ *    type: "thumbnail"
+ *  }
+ * ```
+ */
+export interface ClickedArtistSeriesGroup extends ClickedEntityGroup {
+  action: ActionType.clickedArtistSeriesGroup
```
eessex, line 64: Because this event always has destination `OwnerType.artistSeries`, we can include here to enforce in schema.

src/Schema/Events/SavesAndFollows.ts
```diff
@@ -0,0 +1,68 @@
...
+import { ContextModule } from "../Values/ContextModule"
+import { ScreenOwnerType } from "../Values/OwnerType"
+
+/**
+ * Schemas describing save and follow events
+ * @packageDocumentation
+ */
+
+/**
+ * A user has followed an artist on iOS.
```
eessex, line 11: This looks like it could be generically shared between iOS and web.

src/Schema/Events/Tap.ts
abhitip: Great point! Changing to `curation_boost`

src/Schema/Events/Tap.ts
```diff
@@ -204,7 +205,7 @@ export interface TappedEntityGroup {
   destination_screen_owner_type: ScreenOwnerType
   destination_screen_owner_id?: string
   destination_screen_owner_slug?: string
-  featured?: boolean
+  curation_boost?: boolean
```
eessex, line 208: Does this field apply to all entity types?

abhitip:
Made some updates based off @eessex 's comments. One question for @mikehrom and @anipetrov -

Eve and I discussed how to handle events that would be shared across web and iOS and we decided that we should have just one schema that is used across both. This means that events like `followedArtist` and `unfollowedArtist` would have the same schema in web and iOS but would be stored in their respective tables (force_production.followed_artist vs. eigen_production.followed_artist). What should we name the context_X_owner_type/id/slug properties? I was thinking we just keep it at `context_page_owner_type` etc for simplicity but Eve suggests we drop page/screen and make the properties `context_owner_type`, `context_owner_id` and `context_owner_slug`. Both solutions have the drawback of inconsistency with the rest of the schema for web and iOS but maybe Eve's suggestion is _less_ inconsistent?

src/Schema/Events/Tap.ts
abhitip: It could! But currently we only see this for artistSeries and artwork groups.

mikehrom:
@abhitip if we drop page/screen from the property names, do we have to do any kind of migration on existing data? i do like Eve's suggestion, even though we would need to go through other events and make the same change at some point.

abhitip:
@mikehrom No I don't think we will have to do any migration. I also don't think we'd have to necessarily go through and edit all events. I think it would be `context_page_` for web `context_screen_` for iOS and `context_` for events that are used for both.

I think I'm also being convinced that this is the move!

src/Schema/Events/SavesAndFollows.ts
```diff
@@ -0,0 +1,68 @@
...
+ *    owner_type: "artist"
+ *    owner_id: "5359794d1a1e86c3740001f7"
+ *    owner_slug: "alex-katz"
+ *  }
+ * ```
+ */
+export interface FollowedArtist {
+  action: ActionType.followedArtist
+  context_module: ContextModule
+  context_owner_type: PageOwnerType
```
eessex, line 32: just `OwnerType`

mikehrom:
this looks good! 

we discussed having to potentially coalesce (for example) context_owner_type, context_screen_owner_type, and context_page_owner_type downstream in fulcrum because of these changes, but we think right now, prioritizing a clean rebuild of historic event data with new event property names is unlikely to occur. perhaps later on we work with platform/data engineering on a documented, streamlined process for applying new event properties to existing data.

artsyit:
:rocket: PR was released in [`v1.29.0`](https://github.com/artsy/cohesion/releases/tag/v1.29.0) :rocket:

PR #95, opened 2020-08-20
Use auth context for follow

eessex:
Follow events require the same context as authentication, because it triggers an authentication modal for anonymous users.


artsyit:
:rocket: PR was released in [`v1.31.0`](https://github.com/artsy/cohesion/releases/tag/v1.31.0) :rocket:

PR #96, opened 2020-08-24
Add helpers for ClickedArtworkGroup

eessex:
This PR also serves as a proposal for approaching our `ClickedEntityGroup` helpers a bit differently. All entity groups share the same schema, but the inclusion of events related to header links in these tables forces important fields like `destination_page_owner_id` to be optional. 

Initially we started with a shared helper, `clickedEntityGroup` when implementing the events in web (there is a similar `tappedEntityGroup` too). Even though the schema is shared, by using a shared helper we prevent adding defaults for known values for many of the rails.

This PR introduces two new helpers, `clickedArtworkGroup` and `clickedArtworkGroupHeader`.  By separating the header concept from the artwork click, we are able to make all arguments required.


src/Helpers/Click/ClickedArtworkGroup.ts
```diff
@@ -0,0 +1,107 @@
...
+ *   contextPageOwnerSlug: "andy-warhol-skull",
+ *   contextPageOwnerType: OwnerType.artwork,
+ *   destinationPageOwnerId: "5359794d1a1e86c3740001f7",
+ *   destinationPageOwnerSlug: "andy-warhol-skulls",
+ *   destinationPageOwnerType: OwnerType.artistSeries,
+ *   type: "viewAll",
+ * })
+ * ```
+ */
+export const clickedArtworkGroupHeader = ({
```
sweir27, line 86: Ah! Interesting-- I think currently the schema differentiates between the header and the individual artwork via the `type` param (which seems elegant)! @abhitip -- curious about the decision to break these out?

sweir27: Ooh jk I see now that it's a different `action`. Ignore!

pvinis (approved):

artsyit:
:rocket: PR was released in [`v1.55.0`](https://github.com/artsy/cohesion/releases/tag/v1.55.0) :rocket:

PR #105, opened 2020-09-25
[AS-1555] Analytics tracking for Fairs work on iOS

abhitip:
I've added all the events for iOS in the spreadsheet - https://docs.google.com/spreadsheets/d/1_VDAmRJz2K87FXDq7oh_tQSgwBv4syuZjaaR3sfetcI/edit#gid=849940367

@louislecluse my only question for you is if you think `tappedView` is an okay name for tapping on a tab. We have these tabs on the artist pages too, so I know that we'll be seeing more of these tabs throughout the ecosystem. I could also do `tappedTab` or `tappedNewView`, but there is already a `tappedTabBar` event for the icons at the bottom of the app.


louislecluse:
Same problem for web, I looked at the currently firing event and it is "click" with context_module: NavigationTabs.
So I am creating a clickedNavigationTab event.

I think tappedView makes sense for iOS!

abhitip:
Ooo good call - I'm going to change it to `tappedNavigationTab` actually!

artsyit:
:rocket: PR was released in [`v1.37.0`](https://github.com/artsy/cohesion/releases/tag/v1.37.0) :rocket:

PR #115, opened 2020-10-22
Shows Refresh Tracking

louislecluse:
# What
Adding tracking for the shows refresh on Web and iOS.
Tracking spreadsheet: [link](https://docs.google.com/spreadsheets/d/1_VDAmRJz2K87FXDq7oh_tQSgwBv4syuZjaaR3sfetcI/edit?usp=sharing).

# Discussion:
- There are 3 events that are just clicks on either Viewing Room, Fair, Gallery card. I didn't create specific events for those - I left these as "tap" since it's just a tap on an isolated object. But happy to discuss.

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -210,7 +210,7 @@ export interface ClickedEntityGroup {
 }
 
 /**
- * A user clicks a fair card on a show page
+ * A user clicks a fair card
  *
  * This schema describes events sent to Segment from [[ClickedFairCard]]
  *
@@ -304,7 +304,7 @@ export interface ClickedNavigationTab {
 }
 
 /**
- * A user clicks a partner card on a show page
+ * A user clicks a partner card
  *
  * This schema describes events sent to Segment from [[ClickedPartnerCard]]
  *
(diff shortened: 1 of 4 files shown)


abhitip:
I think you'll need to add a `click` and `tap` event to cohesion because they don't currently exist.

louislecluse:
@abhitip 
Since I would have had to add events, I decided to go the detailed way and add `tappedXCard` events.
"Cards" are being more used, so I suppose we will re-use those events.
Note: Could create a `tappedEntityCard` group as an improvement.

abhitip:
Amazing thank you @louislecluse ! I like the idea of using these events for when there's just one individual item to click on, versus a "group". Do you think it makes sense to update the context_modules to reflect what exactly is being clicked on? Like `showViewingRoom`, `representingGallery`, and `representingFair`?

louislecluse:
Thank you for your comments! Incorporated both!

Going to merge!

artsyit:
:rocket: PR was released in [`v1.45.0`](https://github.com/artsy/cohesion/releases/tag/v1.45.0) :rocket:

PR #117, opened 2020-10-30
[AS-1656] Adds new search events

abhitip:
This PR adds new search events to cohesion + 1 context module needed for the new consign page.
See the /consign2 sheet here - https://docs.google.com/spreadsheets/d/12wj1rt-j5SguboPJMuaM1GqJD-5B5u5piipw8MQ5Zuk/edit#gid=0

@eessex I'm assigning you because I'd like to get your thoughts on this new search file!
@anipetrov let me know if you're good with these search events. They pretty much mirror the current search events for the main search bar, which are not in cohesion yet.

### DIFF
--- a/src/Schema/Events/Search.ts
+++ b/src/Schema/Events/Search.ts
@@ -1,0 +1,89 @@
+import { ActionType } from "."
+import { ContextModule } from "../Values/ContextModule"
+import { OwnerType, PageOwnerType } from "../Values/OwnerType"
+
+/**
+ * Schemas describing Search events
+ * @packageDocumentation
+ */
+
+/**
+ * A user focuses on a search box
+ *
+ * This schema describes events sent to Segment from [[FocusedOnSearchInput]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "focusedOnSearchInput",
+ *    context_module: "header",
+ *    context_page_owner_type: "home"
+ *  }
+ * ```
+ */
+export interface FocusedOnSearchInput {
+  action: ActionType.focusedOnSearchInput
+  context_module: ContextModule
+  context_page_owner_type: PageOwnerType
+  context_page_owner_id?: string
+  context_page_owner_slug?: string
+}
+
+/**
+ * A user selects an item from search
+ *
+ * This schema describes events sent to Segment from [[SelectedItemFromSearch]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "selectedItemFromSearch",
+ *    context_module: "priceEstimate",
+ *    context_page_owner_type: "consign",
+ *    owner_type: "artist",
+ *    owner_id: "5df3e3fa485efe0012c37055",
+ *    owner_slug: "andy-warhol",
+ *    query: "andy warhol"
+ *  }
+ * ```
+ */
+export interface SelectedItemFromSearch {
+  action: ActionType.selectedItemFromSearch
+  context_module: ContextModule
+  context_page_owner_type: PageOwnerType
+  context_page_owner_id?: string
+  context_page_owner_slug?: string
+  destination_page_owner_type?: PageOwnerType
+  destination_page_owner_id?: string
+  destination_page_owner_slug?: string
+  owner_type: OwnerType
+  owner_id: string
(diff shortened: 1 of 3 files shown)


src/Schema/Events/Search.ts
```diff
@@ -0,0 +1,89 @@
...
+ *    action: "focusedOnSearchInput",
+ *    context_module: "header",
+ *    context_page_owner_type: "home"
+ *  }
+ * ```
+ */
+export interface FocusedOnSearchInput {
+  action: ActionType.focusedOnSearchInput
+  context_module: ContextModule
+  context_page_owner_type: PageOwnerType
```
eessex, line 27: Would recommend using generic `context_owner_thing` over page syntax here so we can share these events with iOs.

src/Schema/Events/Search.ts
```diff
@@ -0,0 +1,89 @@
...
+ *  }
+ * ```
+ */
+export interface SelectedItemFromSearch {
+  action: ActionType.selectedItemFromSearch
+  context_module: ContextModule
+  context_page_owner_type: PageOwnerType
+  context_page_owner_id?: string
+  context_page_owner_slug?: string
+  destination_page_owner_type?: PageOwnerType
```
eessex, line 56: `destination_page_owner_type` should likely be required.

src/Schema/Events/Search.ts
```diff
@@ -0,0 +1,89 @@
...
+  destination_page_owner_id?: string
+  destination_page_owner_slug?: string
+  owner_type: OwnerType
```
eessex, line 59: what is the owner type that is neither context nor destination?

src/Schema/Events/Search.ts
```diff
@@ -0,0 +1,89 @@
...
+ *  }
+ * ```
+ */
+export interface SearchedWithNoResults {
+  action: ActionType.searchedWithNoResults
+  context_module: ContextModule
+  context_page_owner_type: PageOwnerType
+  context_page_owner_id?: string
+  context_page_owner_slug?: string
+  destination_page_owner_type?: PageOwnerType
```
eessex, line 87: Is it possible for there to be a destination when there are no results? If there is one `destination_page_owner_type` should probably be required.

src/Schema/Events/Search.ts
abhitip: This is specific to the [new consign page](https://www.figma.com/file/I7a7Iy3yCFi7Pj1VguqMFG/Sell-Art-Landing-%5BWeb%5D?node-id=1155%3A0). In the new design, you can search for an artist to view their market insights. It does not take you to a new page; it just displays the result on your existing page, so I added this additional owner info so that we can track what users are choosing.

abhitip: That's why the destination_page_owner_type field is not required!

src/Schema/Events/Search.ts
abhitip: This is also because of the consign page - if there are no results, you won't go to a separate page.

src/Schema/Events/Search.ts
```diff
@@ -0,0 +1,89 @@
...
+ *    action: "focusedOnSearchInput",
+ *    context_module: "header",
+ *    context_owner_type: "home"
+ *  }
+ * ```
+ */
+export interface FocusedOnSearchInput {
+  action: ActionType.focusedOnSearchInput
+  context_module: ContextModule
+  context_owner_type: PageOwnerType
```
eessex, line 27: If this event is shared between mobile and web we can use `context_owner_type: PageOwnerType | ScreenOwnerType`.

eessex (approved):
One minor comment, but should be non-blocking to add this to web.

src/Schema/Events/Search.ts
abhitip: can I just make it `OwnerType` actually?

abhitip:
Merging for now - we can address the screen type when it comes up

artsyit:
:rocket: PR was released in [`v1.48.0`](https://github.com/artsy/cohesion/releases/tag/v1.48.0) :rocket:

PR #118, opened 2020-10-30
[AS-1657] Adds onboarding event

abhitip:
Adding this so that we can add more tracking to the onboarding flow. The events I'd like to track are in the web onboarding sheet here - https://docs.google.com/spreadsheets/d/12wj1rt-j5SguboPJMuaM1GqJD-5B5u5piipw8MQ5Zuk/edit#gid=603678671

@eessex Let me know what you think about the new event. I added it to `Authentication` because it seemed like the best place to put it. I'd like for the data_input to be a list of the items that people select because right now you can select multiple interests at the interests stage. What should I make the type in that case?
@mikehrom I wasn't sure what to name the event or the `data_input` field. Lmk if you have any thoughts!


src/Schema/Events/Authentication.ts
```diff
@@ -160,3 +161,25 @@ export type AuthTrigger = "click" | "timed"
...
+ *    context_module: "onboardingInterests",
+ *    context_page_owner_type: "onboarding",
+ *    data_input: "[learnAboutArt]"
+ *  }
+ * ```
+ */
+export interface OnboardingUserInputData {
+  action: ActionType.onboardingUserInputData
+  context_module: ContextModule
+  context_page_owner_type: PageOwnerType
```
eessex, line 183: Do we do onboarding for mobile? We might want to go with the generic `context_owner_type`

src/Schema/Events/Authentication.ts
```diff
@@ -160,3 +161,25 @@ export type AuthTrigger = "click" | "timed"
...
+ *
+ * This schema describes events sent to Segment from [[OnboardingUserInputData]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "onboardingUserInputData",
+ *    context_module: "onboardingInterests",
+ *    context_page_owner_type: "onboarding",
+ *    data_input: "[learnAboutArt]"
```
eessex, line 176: Can you clarify what the array syntax in this field is? I also wonder if `data_input` needs a more specific title. We use `subject` in some interfaces to display the text from a clicked item, which seems similar.

abhitip: The issue is that users can choose multiple options in onboarding, so I want to record everything they choose. Would that still be a string? 

Here's an example:
Screen_Shot_2020-09-17_at_11 29 25_AM

src/Schema/Events/Authentication.ts
abhitip: Yup it's on mobile too, I can update this.

src/Schema/Events/Authentication.ts
eessex: In that case it would be typed as:
`data_input: string[]`
An example would read as `data_input: ["learnAboutArt"]` 

Regarding the use of strings for the value, do you want the exact text or a shorthand version? `Learn about art` vs `learnAboutArt` -- if we know to expected values we may want to type the allowed strings.

abhitip: Good question - let's keep it as the exact string actually. I'll update it

eessex:
This schema needs to be added to the `Events` list here: https://github.com/artsy/cohesion/blob/master/src/Schema/Events/index.ts#L54

abhitip:
Done @eessex !

artsyit:
:rocket: PR was released in [`v1.60.0`](https://github.com/artsy/cohesion/releases/tag/v1.60.0) :rocket:

PR #119, opened 2020-11-03
Track sale_screen_load_complete

mikehrom:
[tagged @brainbicycle because we've worked on cohesion event implementation before, and we'll need to do tests/helpers]

- tracks an event for when the sale home screen loads on iOS

### DIFF
--- a/src/Schema/Events/System.ts
+++ b/src/Schema/Events/System.ts
@@ -1,6 +1,5 @@
 import { ActionType } from "."
-import { PageOwnerType } from "../Values/OwnerType"
-import { ScreenOwnerType } from "../Values/OwnerType"
+import { PageOwnerType, ScreenOwnerType } from "../Values/OwnerType"
 
 /**
  * Schemas describing system events
--- a/src/Schema/Events/System.ts
+++ b/src/Schema/Events/System.ts
@@ -27,3 +27,31 @@ export interface TimeOnPage {
   context_page_owner_id?: string
   context_page_owner_slug?: string
 }
+
+/**
+ * A user loads a sale screen on the iOS app.
+ *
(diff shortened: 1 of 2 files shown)


src/Schema/Events/System.ts
```diff
@@ -27,3 +27,31 @@ export interface TimeOnPage {
...
+ *    context_page_owner_slug: "finarte-modern-and-contemporary-art-8",
+ *    jump_to_lai_interface: true,
+ *    load_time_ms: 2000,
+ *    number_of_lots: 276
+ *  }
+ * ```
+ */
+export interface SaleScreenLoadComplete {
+  action: ActionType.saleScreenLoadComplete
+  context_page_owner_type: PageOwnerType
```
eessex, line 51: If this is iOs specific we should use `context_screen_owner_`... and `ScreenOwnerType`.

mikehrom: ah good catch! i will update

src/Schema/Events/System.ts
```diff
@@ -1,5 +1,6 @@
 import { ActionType } from "."
 import { PageOwnerType } from "../Values/OwnerType"
+import { ScreenOwnerType } from "../Values/OwnerType"
```
eessex, line 3: Can use a single import here:
`import { PageOwnerType , ScreenOwnerType } from "../Values/OwnerType"`

mikehrom: 👍

pepopowitz (approved):
Looks good to me!

artsyit:
:rocket: PR was released in [`v1.46.0`](https://github.com/artsy/cohesion/releases/tag/v1.46.0) :rocket:

PR #120, opened 2020-11-04
Track push notification toggles

mikehrom:
In advance of some pretty serious marketing initiatives, I thought it would be useful to track push notification option toggles from our user profile page in the app. This PR adds:

- a screen/page owner type of 'profile' to get started — we'll eventually need to revisit tracking of that screen, too
- push notification types
- new context module for the push notifications screen within profile
- a whole new toggledNotification action 

Let me know if this makes sense!


src/Schema/Values/OwnerType.ts
```diff
@@ -97,6 +99,7 @@ export type PageOwnerType =
   | OwnerType.gene
   | OwnerType.home
   | OwnerType.partner
+  | OwnerType.profile
```
eessex, line 102: We may not need this in `PageOwnerType` unless there is a corresponding Force `profile`.

mikehrom: Yeah, I'm not sure about what happens with web's equivalent of iOS's profile screen and sub-screens. For example at `https://www.artsy.net/profile/edit`: 

Screen Shot 2020-11-05 at 6 28 06 PM

mikehrom: I can remove it for now, if that's preferable.

artsyit:
:rocket: PR was released in [`v1.55.1`](https://github.com/artsy/cohesion/releases/tag/v1.55.1) :rocket:

PR #124, opened 2020-11-09
feat: added tapped register to bid helper

MounirDhahri:
- Added Tapped register to bid helper

### DIFF
--- a/src/Helpers/Tap/TappedRegisterToBid.ts
+++ b/src/Helpers/Tap/TappedRegisterToBid.ts
@@ -19,9 +19,9 @@ export interface TappedRegisterToBidArgs {
  * ```
  * tappedRegisterToBid({
  *   contextModule: true,
- *   contextScreenOwnerType: "sale",
  *   contextScreenOwnerId: "5359794d2a1e86c3741001f8",
  *   contextScreenOwnerSlug "auction-slug":
+ *   contextScreenOwnerType: "sale",
  * })
  * ```
  */
@@ -35,8 +35,8 @@ export const tappedRegisterToBid = ({
   return {
     action: ActionType.tappedRegisterToBid,
     context_module: contextModule,
-    context_screen_owner_type: contextScreenOwnerType,
     context_screen_owner_id: contextScreenOwnerId,
     context_screen_owner_slug: contextScreenOwnerSlug,
+    context_screen_owner_type: contextScreenOwnerType,
   }
 }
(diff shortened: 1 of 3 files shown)


src/Helpers/Tap/__tests__/TappedRegisterToBid.test copy.ts
```diff
@@ -0,0 +1,22 @@
...
+      contextScreenOwnerSlug: "auction-slug",
+      contextScreenOwnerType: OwnerType.sale,
+    })
+
+    expect(event).toEqual({
+      action: "tappedRegisterToBid",
+      contextModule: "aboutThisAuction",
+      contextScreenOwnerId: "5359794d2a1e86c3741001f8",
+      contextScreenOwnerSlug: "auction-slug",
+      contextScreenOwnerType: "sale",
```
pepopowitz, line 19:
```suggestion
      context_module: "aboutThisAuction",
      context_screen_owner_id: "5359794d2a1e86c3741001f8",
      context_screen_owner_slug: "auction-slug",
      context_screen_owner_type: "sale",
```

I think this will fix the broken test?

pepopowitz (approved):
Looks good, I think the broken test is a result of using camelCase instead of snake_case in the assertion. 

I'll reassign to you to merge once you get that cleaned up.

src/Helpers/Tap/TappedRegisterToBid.ts
```diff
@@ -0,0 +1,42 @@
...
+  contextScreenOwnerType,
+  contextScreenOwnerId,
+  contextScreenOwnerSlug,
+}: TappedRegisterToBidArgs): TappedRegisterToBid => {
+  return {
+    action: ActionType.tappedRegisterToBid,
+    context_module: contextModule,
+    context_screen_owner_id: contextScreenOwnerId,
+    context_screen_owner_slug: contextScreenOwnerSlug,
+    context_screen_owner_type: contextScreenOwnerType,
```
eessex, line 40: Does this event occur on multiple owner types? if it is only `sale` pages we may be able to return that by default.

src/Helpers/Tap/__tests__/TappedRegisterToBid.test copy.ts
MounirDhahri: my bad 😬 I forgot to test locally. Thanks 🙇‍♂️

src/Helpers/Tap/TappedRegisterToBid.ts
MounirDhahri: yes, it actually occurs on more than one page. So far sale and sale information pages.

artsyit:
:rocket: PR was released in [`v1.51.0`](https://github.com/artsy/cohesion/releases/tag/v1.51.0) :rocket:

PR #127, opened 2020-11-09
Filter and sort tracking

mikehrom:
I'll leave some comments inline, but I've intended for this to be the main PR for moving our `commercial_filter_params_changed` event into cohesion, for web and iOS. The main issue I need to solve is how to get the filter modal options from [here](https://github.com/artsy/eigen/blob/master/src/lib/Components/FilterModal/FilterModal.tsx) into this event. Ideally when filters are changed or updated, we store JSON in both `current` and `changed`, the latter of which is the updated version of the selected filters. 

For example, I've updated the price range filter from zero state to '1000-5000' in this trace screenshot: 
IMG_6177


src/Schema/Events/FilterAndSort.ts
```diff
@@ -0,0 +1,32 @@
...
+ *  }
+ * ```
+ */
+export interface CommercialFilterParamsChanged {
+  action: ActionType.commercialFilterParamsChanged
+  context_module: ContextModule
+  context_owner_type: OwnerType
+  context_owner_id?: string
+  context_owner_slug?: string
+  current: FilterParams
```
mikehrom, line 30: don't know if this needs to be `JSON` or a store of new Values, but I've started creating a place for said values.

src/Schema/Values/FilterParams.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Master list of filter params available in web and iOS
+ *
+ * Intended action by a user that triggered an event
+ * @packageDocumentation
+ */
+export enum FilterParams {
```
mikehrom, line 7: Here, I want to have all the filters that can possibly change (like `dimensionRange`, `priceRange` etc.) but I am wondering if I also need to have some list of value that those filters can be.

eessex: `FilterParams` should be an `interface` and define the types of content that are allowed, for example `string` or `boolean`.

We can also add an enum for values such as `medium` which would provide a master list of filer options for that key.

eessex (approved):
👍

artsyit:
:rocket: PR was released in [`v1.56.0`](https://github.com/artsy/cohesion/releases/tag/v1.56.0) :rocket:

PR #129, opened 2020-11-09
feat: Add /Consign tracking helpers + eslint autofixers

damassi:
Related: https://github.com/artsy/force/pull/6610 

First cohesion PR! 

This is a follow-up to https://github.com/artsy/cohesion/pull/117 which adds some tracking helpers based on [this tracking matrix](https://docs.google.com/spreadsheets/d/12wj1rt-j5SguboPJMuaM1GqJD-5B5u5piipw8MQ5Zuk/edit#gid=0). 

I'm following the `Helpers/AppName` pattern for one-stop access which (mostly) matches 1:1 to spreadsheet. Assigning to you @eessex as I'm not entirely sure I'm doing this correctly. 

I've also updated our ESlint config with two auto-fixers -- [for imports](https://www.npmjs.com/package/eslint-plugin-simple-import-sort) as well as [object keys](https://www.npmjs.com/package/eslint-plugin-sort-keys-fix) -- so that the developer experience is a bit smoother, and added a `.vscode/settings.json` folder so that all of this is automatic and enforced.


src/Helpers/Consign/index.ts
```diff
@@ -0,0 +1,113 @@
...
+  props: ExcludeDefined<ClickedAppDownload>,
+): ClickedAppDownload => {
+  return {
+    action: ActionType.clickedAppDownload,
+    context_module: ContextModule.sellFooter,
+    context_page_owner_id: props.context_page_owner_id,
+    context_page_owner_slug: props.context_page_owner_slug,
+    context_page_owner_type: OwnerType.consign,
+    destination_path: props.destination_path,
+    subject: "Download the app",
```
damassi, line 111: Should we add a `Subject` enum?

jonallured (approved):
nice one!

src/Helpers/Consign/index.ts
eessex: Usually `subject` is a free-text string pulled in from the button

src/Helpers/Consign/index.ts
```diff
@@ -0,0 +1,113 @@
...
+    context_module: ContextModule.artworkRecentlySoldGrid,
+    context_page_owner_type: OwnerType.consign,
+    destination_page_owner_id: props.destination_page_owner_id,
+    destination_page_owner_slug: props.destination_page_owner_slug,
+    destination_page_owner_type: props.destination_page_owner_type,
+    type: "thumbnail",
+  }
+}
+
+export const clickedArticleGroup = (
```
eessex, line 87: This is a click event that is not specific to consign contexts and should live in the `Click` folder: https://github.com/artsy/cohesion/tree/master/src/Helpers/Click

src/Helpers/Consign/index.ts
```diff
@@ -0,0 +1,113 @@
...
+    context_page_owner_type: OwnerType.consign,
+    destination_page_owner_id: props.destination_page_owner_id,
+    destination_page_owner_slug: props.destination_page_owner_slug,
+    destination_page_owner_type: props.destination_page_owner_type,
+    type: "thumbnail",
+  }
+}
+
+export const clickedAppDownload = (
+  props: ExcludeDefined<ClickedAppDownload>,
```
eessex, line 102: Same here, should be considered a "click" event rather than consign-specific.

src/Helpers/Consign/index.ts
```diff
@@ -0,0 +1,113 @@
...
+    context_module: ContextModule.priceEstimate,
+    context_owner_id: props.context_owner_id,
+    context_owner_slug: props.context_owner_slug,
+    context_owner_type: OwnerType.consign,
+    destination_owner_type: props.destination_owner_type,
+    query: props.query,
+  }
+}
+
+export const clickedArtworkGroup = (
```
eessex, line 73: Just merged a pre-existing PR for this event helper, can be removed.

src/Helpers/Consign/index.ts
```diff
@@ -0,0 +1,113 @@
...
+
+type ExcludeDefined<T> = Required<
+  Omit<
+    T,
+    "action" | "context_module" | "context_owner_type" | "type" | "subject"
+  >
+>
+
+export const focusedOnSearchInput = (
+  props: ExcludeDefined<FocusedOnSearchInput>,
```
eessex, line 29: Search events should live together in a `Helpers/Search` directory.

eessex:
Some minor changes related to placement of events in the file tree-- we share events across all page-types, and sometimes between iOS and web, so all schema/helpers are arranged by the type of event (click, search, authentication, etc) rather then where in the app they get triggered.

.eslintrc.js
```diff
@@ -16,12 +16,18 @@ module.exports = {
     ecmaVersion: 6,
     sourceType: "module",
   },
-  plugins: ["@typescript-eslint", "jest", "eslint-plugin-import"],
+  plugins: [
+    "@typescript-eslint",
```
eessex, line 20: 👍

package.json
```diff
@@ -39,6 +39,8 @@
     "eslint-config-prettier": "6.11.0",
     "eslint-plugin-import": "2.20.2",
     "eslint-plugin-jest": "23.13.2",
+    "eslint-plugin-simple-import-sort": "^5.0.3",
```
eessex, line 42: 🙏

src/Helpers/Consign/index.ts
damassi: Cool, so in that case we pass in below hard-coded values as props and this function becomes reusable 👍

williardx (approved):
LGTM!

src/Helpers/Click/ClickedAppDownload.ts
```diff
@@ -0,0 +1,15 @@
+import { ActionType, ClickedAppDownload } from "../../Schema"
+
+export const clickedAppDownload = (
```
eessex, line 3: If there is no property transformation happening in the helper you can probably just type the tracking props via the interface, this won't accomplish much besides saving us from passing the `action` field.

src/Helpers/Click/ClickedArtworkGroup.ts
```diff
@@ -9,8 +9,8 @@ import {
 export interface ClickedArtworkGroupArgs {
   contextModule: ContextModule
   contextPageOwnerType: PageOwnerType
-  contextPageOwnerId: string
-  contextPageOwnerSlug: string
+  contextPageOwnerId?: string
```
eessex, line 12: Which page is this being used on that has no id/slug?

src/Helpers/Click/ClickedAppDownload.ts
damassi: This is useful because we have quite a few links around the site where we download the app, now we can just import this helper when we need to.

src/Helpers/Click/ClickedArtworkGroup.ts
damassi: artsy.net/consign

damassi:
@eessex - going to merge so that we can get new consign page into QA. Looking over the new helpers, they do seem generic / reusable enough even if they don't have any logic. For example the search stuff could all apply to top search bar, or any other filter interface, etc. Happy to follow up and remove if you're adamant about it tho! (Or any other feedback you might find)

artsyit:
:rocket: PR was released in [`v1.58.0`](https://github.com/artsy/cohesion/releases/tag/v1.58.0) :rocket:

src/Helpers/Click/ClickedAppDownload.ts
eessex: This seems to just duplicate the `ClickedAppDownload` schema as a function-- I don't see the utility, also breaks the props transformation pattern from the existing helpers?

PR #131, opened 2020-11-12
Track add_to_calendar event

mikehrom:
Adds tracking for when a user clicks 'Add to Calendar' for time-based events.


src/Schema/Events/Click.ts
```diff
@@ -418,3 +418,29 @@ export interface ClickedAppDownload {
...
+ *  }
+ * ```
+ */
+export interface ClickedAddToCalendar {
+  action: ActionType.clickedAddToCalendar
+  context_module: ContextModule
+  context_page_owner_type: PageOwnerType
+  context_page_owner_id?: string
+  context_page_owner_slug?: string
+  subject: "Google" | "Outlook" | "iCal"
```
mikehrom, line 445: copied the case exactly for these, but should they be all lower-case? can update if needed

src/Schema/Events/Click.ts
```diff
@@ -418,3 +418,29 @@ export interface ClickedAppDownload {
...
+ *    action: "clickedAddToCalendar",
+ *    context_module: "auctionHome",
+ *    context_page_owner_type: "sale",
+ *    context_page_owner_id: "5f99e0ba4c24bc000d02b8d7",
+ *    context_page_owner_slug: "artsy-x-seoul-auction-contemporary-icons",
+ *    subject: "google"
+ *  }
+ * ```
+ */
+export interface ClickedAddToCalendar {
```
eessex, line 439: Considering that we also have this feature in iOs, this could be a good candidate to make agnostic to `click` or `tap` actions and go with a more generic `AddToCalendar`, similar to how the `Follow` event is setup: https://cohesion.artsy.net/interfaces/_schema_events_savesandfollows_.followedartist.html 

For events that can be shared between apps we are also using the generic `context_owner_type` rather than `context_screen`... or `context_page`... 

cc @abhitip

src/Schema/Events/Click.ts
eessex: Consistency sounds good to me, `google` / `outlook` - `iCal` still matches camelCase convention.

src/Schema/Events/Click.ts
abhitip: I don't think that's a bad idea! The events will still end up in separate schemas/tables in Redshift `force_production` vs `eigen_production`. Making it one consistent schema will make it easier to analyze. What do you think @mikehrom?

mikehrom: Agree 👍  will update

mikehrom:
@abhitip @eessex I'm ready for another review after making this platform-agnostic 👍

artsyit:
:rocket: PR was released in [`v1.59.0`](https://github.com/artsy/cohesion/releases/tag/v1.59.0) :rocket:

PR #134, opened 2020-11-23
Add new events and contexts for auction results on iOS

mikehrom:
Introducing: event tracking schema for the bulk of native auction results features headed to iOS. Information bubble taps are not included here; I'll follow up with those later as they are not critical or a blocker.

### DIFF
--- a/src/Schema/Events/FilterAndSort.ts
+++ b/src/Schema/Events/FilterAndSort.ts
@@ -37,7 +37,26 @@ export interface CommercialFilterParams {
 }
 
 /**
- * A user applies filters to a filterable/sortable module
+ * Master list of filter params available in web and iOS, specific to auction results
+ *
+ * Intended action by a user that triggered an event
+ * @packageDocumentation
+ */
+export interface AuctionResultsFilterParams {
+  allowEmptyCreatedDates?: boolean
+  categories?: string[]
+  createdAfterYear?: string
+  createdBeforeYear?: string
+  earliestCreatedYear?: string
+  latestCreatedYear?: string
+  organizations?: string[]
+  pageAndCursor?: string[]
+  sizes?: string[]
+  sort?: string
+}
+
+/**
+ * A user applies filters to a filterable/sortable artworks module
  *
  * This schema describes events sent to Segment from [[CommercialFilterParamsChanged]]
  *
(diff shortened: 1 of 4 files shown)


src/Schema/Events/FilterAndSort.ts
```diff
@@ -37,7 +37,26 @@ export interface CommercialFilterParams {
 }
 
 /**
- * A user applies filters to a filterable/sortable module
+ * Master list of filter params available in web and iOS, specific to auction results
+ *
+ * Intended action by a user that triggered an event
+ * @packageDocumentation
+ */
+export interface AuctionResultsFilterParams {
```
mikehrom, line 45: this list may not be exhaustive, but I generated the filter params based on inspecting segment tracks when filtering auction results on web. we can collaborate on which ones need to be added, if any. They look a bit different from the CommercialFilter params.

src/Schema/Events/Tap.ts
```diff
@@ -139,6 +139,29 @@ export interface TappedAuctionGroup extends TappedEntityGroup {
...
+ *  @example
+ *  ```
+ *  {
+ *    action: "tappedAuctionResultGroup",
+ *    context_module: "auctionResults",
+ *    context_screen_owner_type: "artistInsights",
+ *    context_screen_owner_id: "51aa03df8b3b8177260002ab",
+ *    context_screen_owner_slug: "nicolas-party",
+ *    destination_screen_owner_type: "auctionResult",
+ *    destination_screen_owner_id: "6398282",
```
mikehrom, line 156: This assumes that an auction result screen has an auction result id in `diffusion.lots`.

src/Schema/Events/FilterAndSort.ts
```diff
@@ -63,3 +82,31 @@ export interface CommercialFilterParamsChanged {
   current: CommercialFilterParams
   changed: CommercialFilterParams
 }
+
+/**
+ * A user applies filters to a filterable/sortable auction results module
```
eessex, line 87: Do we definitely want a separate event from `CommercialFilterParamsChanged`? The existing one could handle this by making `current`/`changed` accept `AuctionResultsFilterParams | CommercialFilterParams`.

mikehrom: That's a good point. One event, separate params. I did this mainly because right now on web it _is_ a separately named event altogether, but we can certainly go with a single event.

eessex (approved):
Minor q otherwise this looks 👍 to me. It looks like we don't have strict typing on the Force side for filter params, but maybe iOS has an existing schema that can be referenced against.

src/Schema/Events/FilterAndSort.ts
mikehrom: Like this? 
Screen Shot 2020-11-23 at 1 11 35 PM

mikehrom: And then we subsequently remove the references to `auctionResultsFilterParamsChanged` throughout

MounirDhahri (approved):

MounirDhahri:
looking great 🎉

artsyit:
:rocket: PR was released in [`v1.62.0`](https://github.com/artsy/cohesion/releases/tag/v1.62.0) :rocket:

PR #139, opened 2020-11-30
Track critical myCollection events and contexts

mikehrom:
- groups add, edit, delete and tap events for the My Collection flow
- adds contexts and owners for myCollection and myCollectionArtwork
- specs [here](https://www.figma.com/file/yDSHlbJVzBxvSUROD2bfRI/Sell-tab?node-id=1%3A2) and [here](https://www.figma.com/file/yDSHlbJVzBxvSUROD2bfRI/Sell-tab?node-id=466%3A24999)

### DIFF
--- a/src/Schema/Events/MyCollection.ts
+++ b/src/Schema/Events/MyCollection.ts
@@ -26,6 +26,7 @@ export interface AddCollectedArtwork {
   context_module: ContextModule.myCollectionHome
   context_owner_type: OwnerType.myCollection
   context_owner_id?: string
+  context_owner_slug?: string
 }
 
 /**
@@ -52,6 +53,7 @@ export interface EditCollectedArtwork {
   context_module: ContextModule.myCollectionArtwork
   context_owner_type: OwnerType.myCollectionArtwork
   context_owner_id?: string
+  context_owner_slug?: string
 }
 
 /**
@@ -78,6 +80,7 @@ export interface DeleteCollectedArtwork {
   context_module: ContextModule.myCollectionArtwork
   context_owner_type: OwnerType.myCollectionArtwork
   context_owner_id?: string
+  context_owner_slug?: string
 }
 
 /**
@@ -100,5 +103,8 @@ export interface TappedCollectedArtwork {
   context_module: ContextModule.myCollectionHome
   context_owner_type: OwnerType.myCollection
   context_owner_id?: string
+  context_owner_slug?: string
   destination_owner_type: OwnerType.myCollectionArtwork
+  destination_owner_id?: string
+  destination_owner_slug?: string
 }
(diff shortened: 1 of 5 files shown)


src/Schema/Events/MyCollection.ts
```diff
@@ -0,0 +1,104 @@
...
+ *    context_module: "myCollectionHome",
+ *    context_owner_type: "myCollection"
+ *  }
+ * ```
+ */
+export interface AddCollectedArtwork {
+  action: ActionType.addCollectedArtwork
+  context_module: ContextModule.myCollectionHome
+  context_owner_type: OwnerType.myCollection
+  context_owner_id?: string
```
eessex, line 28: We usually collect `slug` as well as `id` when applicable, why not in this case?

mikehrom: good q. i don't know if collections in 'My Collection' will even have slugs, but I assume they will have an id, which i left optional.

mikehrom: could throw it in as `?` optional to be consistent

eessex: Most records with an `id` should also have a slug, but not sure of the details for an individual user's collection. I do wonder if we want to know the `owner_id` and `owner_slug` of the collected artwork?

mikehrom: i left details about the collected artwork off of this event because it's possible a user can tap 'add artwork' and then exit the flow before submitting, but i suppose in those instances, 'destination' info would be null, while for completed submissions, the destination info would be present.

mikehrom: actually, i take that back. the event will fire before we know anything about the added artwork, so i think in this case, we won't need 'destination' info.

mikehrom: (there should be a backend event recorded for a submitted artwork to My Collection)

mikehrom:
Screen Shot 2020-11-30 at 12 56 53 PMScreen Shot 2020-11-30 at 12 57 11 PM

For these highlighted links, I'm wondering if we can use TappedShowMore with the appropriate subject, context, and destination information. Feels unideal to make different events for each, though I submit the blue-boxed button is a bit unique in that it directs users to the current sell tab flow. What do you guys think? @samchieng @brainbicycle

samchieng:
@mikehrom i would agree with that, for the red boxes i think we just want to get a rough idea if users are exploring more info on this page—so don't need to make different events for each. But separating out the blue box would give us insight into potential sell conversion.

brainbicycle (approved):

brainbicycle:
LGTM, feel free to merge when ready! Not sure if you are waiting on any other feedback

artsyit:
:rocket: PR was released in [`v1.64.0`](https://github.com/artsy/cohesion/releases/tag/v1.64.0) :rocket:

PR #141, opened 2020-12-07
Tracked events for inbox on iOS

mikehrom:
For tapping modules on the 'active bids' and 'closed bids' tabs (inboxBids screen), should we use existing tracks like 'TappedArtworkGroup' and 'TappedAuctionGroup' with descriptive contexts? Or should we use a new event called "TappedInboxGroup" or "TappedActiveLot" (naming tbd)? still thinking, but wanted to put this up in the meantime while we discuss and work through.


src/Schema/Values/OwnerType.ts
```diff
@@ -73,7 +75,9 @@ export type ScreenOwnerType =
   | OwnerType.fair
   | OwnerType.fairArtworks
   | OwnerType.home
-  | OwnerType.inbox
```
mikehrom, line 76: I think we should fire different screen events with a context and owner for the 'Inquiries' view and the 'Bids' view within inbox. This is a unique pattern AFAIK but allows us to group all screen views as type 'inbox' at the session level while still being able to distinguish between views if need be.

src/Schema/Values/Tab.ts
```diff
@@ -5,10 +5,9 @@
...
 
 export type Tab =
-  | "cityGuide"
-  | "favorites"
-  | "home"
-  | "messages"
-  | "profile"
-  | "search"
-  | "sell"
+  /**
```
mikehrom, line 8: Quick note here, tapping 'inbox' icon on the footer tab navigation currently fires with tab type = "messages" but I find that naming inferior to "inbox" -- so data team will have to change historic track data, but I think it's worthwhile.

mikehrom:
Screen Shot 2020-12-07 at 7 25 40 PM

- green box
  - considering using [TappedNavigationTab](https://github.com/artsy/cohesion/blob/8361a1d5b13ddc0b3476d5d4e284bee505e1b696/src/Schema/Events/Tap.ts#L529). the reason i'm not sure is because this is highest-level, screen-to-screen navigation and we do not have a similar pattern in the app to date. @eessex curious if you think this deserves its own event.
- pink box
  - use [TappedNavigationTab](https://github.com/artsy/cohesion/blob/8361a1d5b13ddc0b3476d5d4e284bee505e1b696/src/Schema/Events/Tap.ts#L529). This is the established use case for this event, as seen on fair pages.
- yellow box
  - suggesting [TappedAuctionGroup](https://github.com/artsy/cohesion/blob/8361a1d5b13ddc0b3476d5d4e284bee505e1b696/src/Schema/Events/Tap.ts#L118) or TappedAuctionCard, which would be a new event structured similarly to [TappedPartnerCard](https://github.com/artsy/cohesion/blob/8361a1d5b13ddc0b3476d5d4e284bee505e1b696/src/Schema/Events/Tap.ts#L554). What do you think? It's not like a home screen auctions rail, so perhaps Card is more appropriate here. These should navigate to a sale home page.
- blue box
  - suggesting [TappedArtworkGroup](https://github.com/artsy/cohesion/blob/8361a1d5b13ddc0b3476d5d4e284bee505e1b696/src/Schema/Events/Tap.ts#L94), since this is a commercial grouping of artworks. Context screen information would correspond to the inbox, destination screen information would correspond to the lot/artwork tapped. Open to including other context about the lot being tapped.

mikehrom:
Screen Shot 2020-12-07 at 7 25 49 PM
- tapping artworks on this list, if they navigate to lot screens will also fire TappedArtworkGroup (context module: inboxClosedBids)

mikehrom:
Screen Shot 2020-12-07 at 7 35 47 PM
- don't have a tap event for this button. i'm not sure we have an existing button, but i could generate a TappedExplore generic event. thoughts?

mikehrom:
Screen Shot 2020-12-07 at 7 35 56 PM

- finally, TappedInboxMessage (new in this PR) is for tapping inquiries on the inboxInquiries screen

eessex:
- Would `tappedShowMore` work for the `Explore auctions` CTA button?
- I think the `tappedNavigationTabs` makes sense for the bids/inquiries navigation - although they are visually different, it is still a very literal description of the action.
- I'm a bit confused by the difference between an `entity group` and a `card` -- what distinguishes these?

mikehrom:
@eessex `tappedShowMore` could work for the Explore Auctions button, yes. it's definitely a similar-type of interaction, so I think we could cover it there. I like this solution rather than creating a new, similar event. 

good question on distinguishing cards from entity groups. i believe the first instance of this came from viewing rooms via @louislecluse - Louis, was there a particular reason you made new 'Card' events here? It looks like entity groups and Cards function similarly, but appear differently.

I don't mind using tappedAuctionGroup on this PR, and I do think it fits fine. i suppose i was comparing these to where we have previously used tappedAuctionGroup, which is primarily on auctions rails on the home screen, and thinking the two interactions were maybe different enough to merit unique events.

louislecluse:
I _think_ card came from the pattern where we only have 1 entity to click on.
For shows, a show can be link to 1 Viewing Room and we print the "Viewing Room card" for users to click on. So there is 1 entity in the module.

mikehrom:
I'm going to re-visit tappedInboxMessage, since it appears we have some existing tracking related to _conversations_ and I want to square up the naming.

erikdstock:
It seems like we are mostly at a consensus here, but as I am new to this repo i'm not sure when it's okay to merge. I'm personally ok from here!

eessex (approved):

artsyit:
:rocket: PR was released in [`v1.66.0`](https://github.com/artsy/cohesion/releases/tag/v1.66.0) :rocket:

PR #142, opened 2020-12-08
Track screen views

mikehrom:
- A light tracking schema for screen view events (`eigen_production.screens`)


src/Schema/Events/Screen.ts
```diff
@@ -0,0 +1,29 @@
...
+ *    context_screen_owner_id: "527ac4a0cd530e258d0000d0",
+ *    context_screen_owner_slug: "ramiro-gomez"
+ *  }
+ * ```
+ */
+export interface Screen {
+  action: ActionType.screen
+  context_screen_owner_type: ScreenOwnerType
+  context_screen_owner_id?: string
+  context_screen_owner_slug?: string
```
mikehrom, line 28: can we put referrer information here? referring screen type is something we currently don't have good coverage on

src/Schema/Events/Screen.ts
```diff
@@ -0,0 +1,29 @@
...
+ *    action: "screen",
+ *    context_screen_owner_type: "artist",
+ *    context_screen_owner_id: "527ac4a0cd530e258d0000d0",
+ *    context_screen_owner_slug: "ramiro-gomez"
+ *  }
+ * ```
+ */
+export interface Screen {
+  action: ActionType.screen
+  context_screen_owner_type: ScreenOwnerType
```
mikehrom, line 26: i think we'll have to coalesce `context_screen_owner_type` with `name` in our rollup at `analytics.screens` because of this

src/Schema/Events/Screen.ts
abhitip: That sounds like a good idea!

abhitip (commented):
What are your thoughts on adding this to the `System` file instead of creating a new `Screen` file for it?

eessex:
Just want to note that we may need to tweak the client implementation for whoever implements this to use the `screen` event rather then `track1: https://segment.com/docs/connections/sources/catalog/libraries/mobile/react-native/#screen

artsyit:
:rocket: PR was released in [`v1.67.0`](https://github.com/artsy/cohesion/releases/tag/v1.67.0) :rocket:

PR #147, opened 2020-12-14
chore: add tappedNavigationTab helper

eessex:
Adds helper for `TappedNavigationTab` events.

Adds a minimal helper for use in https://github.com/artsy/eigen/pull/4253/files


src/Helpers/Tap/TappedNavigationTab.ts
```diff
@@ -0,0 +1,28 @@
...
+ * A user taps a navigation tab on iOS
+ *
+ *  @example
+ *  ```
+ *  tappedNavigationTab({
+ *    context_module : "exhibitorsView",
+ *    context_screen_owner_type: "fair",
+ *    context_screen_owner_id: "5f4d80c972737e000deff1ed",
+ *    context_screen_owner_slug: "latitude-art-fair-2020",
+ *    subject: "View more"
```
eessex, line 13: @mikehrom LMK if there's a better example text to use for `subject`.

mikehrom: I think this is fine! In the original use case (fairs) we didn't need a subject, but in the Inbox use case, it will be "Bids" or "Inquiries" -- beyond that, it could be a number of options.

mikehrom (approved):

artsyit:
:rocket: PR was released in [`v1.67.2`](https://github.com/artsy/cohesion/releases/tag/v1.67.2) :rocket:

PR #148, opened 2020-12-14
Track entry points to identity verification

mikehrom:
- tracking for buttons that lead to the identify verification flow, like the one below:
Screen Shot 2020-12-14 at 3 09 18 PM

### DIFF
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -290,7 +290,7 @@ export interface TappedConsign {
 /**
  * A user taps a fair card
  *
- * This schema describes events sent to Segment from [[TappedFairCard]]
+ * This schema describes events sent to Segment from [[tappedFairCard]]
  *
  *  @example
  *  ```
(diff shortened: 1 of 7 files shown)


src/Schema/Events/Tap.ts
```diff
@@ -529,7 +529,7 @@ export interface TappedShowMore {
 /**
  * A user taps a navigation tab on iOS
  *
- * This schema describes events sent to Segment from [[tappedNavigationTab]]
+ * This schema describes events sent to Segment from [[TappedNavigationTab]]
```
eessex, line 532: We want these links to be lowercase, so that they cross-link to the relevant helper in the docs: https://cohesion.artsy.net/interfaces/_schema_events_click_.clickedarticlegroup.html

mikehrom: ah, i think i have several to edit then :)

artsyit:
:rocket: PR was released in [`v1.67.4`](https://github.com/artsy/cohesion/releases/tag/v1.67.4) :rocket:

PR #149, opened 2020-12-14
Add context module for YourActiveBids

mikehrom:
#minor adds a context module for the YourActiveBids rail that is visible on iOS home when the user is engaged in auctions


artsyit:
:rocket: PR was released in [`v1.67.3`](https://github.com/artsy/cohesion/releases/tag/v1.67.3) :rocket:

PR #156, opened 2021-01-12
add clickedShowMore helper

eessex:
The type of this PR is: chore

### Description

Add a minimal helper and example for `clickedShowMore`

- [ ] No platform-specific terminology has been used outside of `click` and `tap` (platform is infered by the DB storing events)

The existing interface uses `context_page_owner...` rather than owner but leaving as is, I believe the columns are pre-existing from the old schema.


artsyit:
:rocket: PR was released in [`v1.68.3`](https://github.com/artsy/cohesion/releases/tag/v1.68.3) :rocket:

PR #157, opened 2021-01-20
feat: adds position and sort args to artwork click

dzucconi:

Re: [FX-2545](https://artsyproduct.atlassian.net/browse/FX-2545)


artsyit:
:rocket: PR was released in [`v1.69.0`](https://github.com/artsy/cohesion/releases/tag/v1.69.0) :rocket:

PR #159, opened 2021-01-28
Share

mikehrom:
The type of this PR is: **Instrumentation**

This PR resolves [CX-832], [AS-1889]

### Description

Creating a `share` event for use in force or eigen [[specs]](https://www.figma.com/file/oI0G7GRmNy0lBztSBczRsV/Instagram-Share?node-id=105%3A1)

### PR Checklist

[CX-832]: https://artsyproduct.atlassian.net/browse/CX-832
[AS-1889]: https://artsyproduct.atlassian.net/browse/AS-1889


src/Schema/Values/Service.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * All available share services
+ *
+ * @packageDocumentation
+ */
+export enum Service {
```
mikehrom, line 6: This list is probably not exhaustive, so we can add to it now if necessary

pvinis:
This seems good for when we share on something. What about if the user cancels the share, or when opening the sharesheet? Do we care?

mikehrom:
I think for now, we can implement without error states/cancel states. I don't anticipate analysis involving them -- even though I do think at some point they might be valuable to track.

For the share sheet, what I mainly want to know is:
1. how many times are users sharing artworks (or other pages)?
2. when they share, which service are they using?

If we can get all of that in one event, that's preferable. If you think pairing would make this clearer to me, let's do that in the coming week. I might be missing something about the flow itself.

pvinis:
Closing in favor of #162.

PR #165, opened 2021-02-17
Adds onboarding owner type

abhitip:
I forgot to add the `onboarding` owner type so here it is!

This PR resolves [CO-]

- [ ] I've added comments with examples for any new interfaces or helpers and ensured that they're in the docs 


artsyit:
:rocket: PR was released in [`v1.74.0`](https://github.com/artsy/cohesion/releases/tag/v1.74.0) :rocket:

PR #166, opened 2021-02-18
chore: add missing screens to OwnerType

MounirDhahri:

### Description
- This PR is a preparation to get all screen tracking to rely on cohesion. Some screens use to start with an upper case when they were being tracked in eigen. But since all screens in cohesion follow a snake case pattern, I adjusted them to be the same. I hope that doesn't affect the data so much if the query you are running is case sensitive.
- Since we will be soon releasing android, I changed `iOS` to `App` everywhere in the repo.


mikehrom:
I think this is all well and good. Tagging @abhitip with a question - what is more important (for our needs) is that we categorize/group these owners [here](https://github.com/artsy/fulcrum/tree/master/dbt/data), right?

mikehrom (approved):

abhitip:
This is awesome. Thank you @MounirDhahri !

Yup @mikehrom that's where we will need to take care of the case-sensitive mappings.

@mikehrom we should also do thorough QA to make sure we don't break anything. [Here's](https://www.notion.so/artsy/How-to-Migrate-existing-events-into-cohesion-3e380f55eb434473a3c24a22527a1116) the notion page outlining how to migrate existing events into cohesion. Feel free to put time on my cal this week so that we can write out what SQL queries we will want to monitor!

abhitip (approved):

PR #168, opened 2021-02-27
[AS-1940] New Checklist Events in Cohesion 

brendamanganelli:
Adds CMS Checklist Events and toDoList context module.

- Added new Context Module: toDoList
- Added following Click events: ClickedEditArtwork, ClickedSnooze, ClickedExpandArrow, ClickedCollapseArrow, ClickedLoadMore

[New Checklist design for reference ](https://www.figma.com/file/N0Xh5DaHaXo8q72a5FtVWh/CMS-Checklist?node-id=128%3A0)

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -420,3 +420,115 @@ export interface ClickedAppDownload {
   destination_path: string
   subject: string
 }
+
+/**
+ * A partner clicks on Edit Artwork from CMS Checklist To-Do Items.
+ *
+ * This schema describes events sent to Segment from [[ClickedEditArtwork]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "clickedEditArtwork",
+ *    context_module: "toDoList",
+ *    context_page_owner_type: "home",
+ *    destination_page_owner_type: "artworks",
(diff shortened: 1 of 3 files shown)


abhitip:
FYI @damassi I think this is our first time adding CMS events into cohesion. It looks like these are all events meant for the CMS checklist. How do you think we should organize them? For context, we have a whole file for `Conversations` events for the inquiry inbox. I think it could be helpful for us if we had all these CMS checklist events in their own file, but I'm curious what's helpful from an engineering perspective.

damassi:
Looking at all of these events, while they're specific to CMS they're also very generic naming-wise, and I can imagine many of them being used in other contexts (at some future point). As long as everything is alphabetical in the file, and is loosly organized around function, I think we should continue regarding cohesion as one giant index, versus breaking it apart into service or feature boundaries. 

I can certainly be persuaded tho! This PR looks ideal IMO, however.

abhitip:
Sounds good, thanks @damassi !

LGTM

artsyit:
:rocket: PR was released in [`v1.76.0`](https://github.com/artsy/cohesion/releases/tag/v1.76.0) :rocket:

PR #169, opened 2021-03-04
[AS-1950] Inquiry checkout events

abhitip:
The type of this PR is: adding new events for inquiry checkout

Schema [here](https://docs.google.com/spreadsheets/d/1NIzJ8jQ1BzGxDDmB0pYWjsmfCk3GRRgxtmjTJV1zWg0/edit#gid=0)!

This PR resolves [CO-]

- [ x] If I've added a new file to the tree I've exported it from the common `index.ts`
- [ x] I've added comments with examples for any new interfaces or helpers and ensured that they're in the docs 
- [ x] No platform-specific terminology has been used outside of `click` and `tap` (platform is inferred by the DB storing events)


src/Schema/Events/Conversations.ts
```diff
@@ -2,6 +2,72 @@ import { ContextModule } from "../Values/ContextModule"
...
+ * This schema describes events sent to Segment from [[acceptOffer]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "acceptOffer",
+ *    context_owner_type: "inboxConversation"
+ *    impulse_conversation_id: "198"
+ *    subject: "Confirm & Pay"
+ *  }
```
starsirius, line 17: [Commented](24c25e5a-ec98-466e-864d-c3dd9962a27b) on the spreadsheet that we decided to reuse the webview so we won't be building the review offer screen. We can probably ignore the `AcceptOffer`, `CounterOffer` and `DeclineOffer` here if we already have tracking for the existing flow.

starsirius: Not add add more work but I wonder if tracking tapping on the order CTA would be worthwhile, such as the green "Offer Accepted" bar.

Screen Shot 2021-03-05 at 2 56 02 PM

abhitip: Yes we want to track that with the `tappedViewOffer` event using the `subject` field to figure out the status that they are tapping on. Does that make sense?

anipetrov:
this looks great! @abhitip i will let you merge this when you align with @starsirius on the event/properties for the view offer banner interaction!

src/Schema/Values/OwnerType.ts
```diff
@@ -72,6 +74,8 @@ export type ScreenOwnerType =
   | OwnerType.collection
   | OwnerType.consignmentSubmission
   | OwnerType.consign
+  | OwnerType.conversation
```
starsirius, line 77: I notice we have [`inboxConversation`](https://github.com/artsy/cohesion/blob/e6a7e2cbfec2e61151caff87224078e26a4ae131/src/Schema/Values/OwnerType.ts#L80) as an owner type. I'm not sure what the difference is but want to raise it.

starsirius (approved):
I have a question about `conversation` and `inboxConversation` owner type; otherwise looks great to me! Thank you for helping with this! @abhitip 🙏

src/Schema/Values/OwnerType.ts
abhitip: Great question! @anipetrov and I discussed `inboxConversation` as well. We decided that ideally this would have been `conversation` from the beginning and it's only being used as the destination for the `tappedInboxConversation` event right now. Since we are now starting to implement events on the actual `conversation` screen, we decided that it's okay to switch to `conversation` moving forward.

Let me know if that's okay!

starsirius: Yeah that makes sense to me! Thanks for the context! 👍

artsyit:
:rocket: PR was released in [`v1.75.0`](https://github.com/artsy/cohesion/releases/tag/v1.75.0) :rocket:

PR #172, opened 2021-03-19
Chore(PX-3846): Clicked[Expand|Collapse]Arrow -> ClickedExpansionToggle

mdole:
The type of this PR is: Enhancement

This PR resolves [PX-3846]

### Description

This PR updates naming and examples for events we expect CMS checklist to consume.

[PX-3846]: https://artsyproduct.atlassian.net/browse/PX-3846

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -435,6 +435,7 @@ export interface ClickedAppDownload {
  *    destination_page_owner_type: "artworks",
  *    destination_page_owner_id: "5808b9a0cd530e658500008a",
  *    destination_page_owner_slug: "maddalena-ambrosio-untitled"
+ *    destination_path: "/artworks/maddalena-ambrosio-untitled/edit"
  *  }
  * ```
  */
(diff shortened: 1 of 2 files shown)


src/Helpers/Click/ClickedEditArtwork.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * A user clicks to edit an artwork on web.
+ *
+ *  @example
+ *  ```
+ *  clickedExpansionToggle({
+ *    contextModule: ContextModule.toDoList,
+ *    contextPageOwnerType: OwnerType.home,
+ *    destinationPageOwnerType: OwnerType.artwork,
```
mdole, line 25: @brendamanganelli question I had while working on this: do you think we need separate OwnerTypes for ArtworkEdit and/or CMSHome since those are distinct from artworks/home on artsy.net?

src/Helpers/Click/ClickedEditArtwork.ts
```diff
@@ -0,0 +1,46 @@
...
+ *    destinationPageOwnerId: "5ff5c57b5022f600116d7a1a",
+ *    destinationPageOwnerSlug: "daniel-heidkamp-el-capitan-night",
+ *    destinationPath: "/artworks/daniel-heidkamp-el-capitan-night/edit"
```
mdole, line 28: And one other question for you @brendamanganelli - I noticed that `destination_path` is in the interface in the schema but not the example, nor is it in the spreadsheet. Do we want it?

damassi:
@mdole - Just responded in that slack thread, but TBH I don't think any of these helpers are necessary! (Need to clarify this with the team.) Since they only remap stuff to snake case and don't do any real computing I think its just a bit of overhead, including 1:2 drift (interface needs snake case, props need camelCase -> divergence = complexity cost). 

So yah I think we can import the interfaces directly into our app and track there. But if you really like the helpers, feel free to merge 😄.

src/Helpers/Click/ClickedEditArtwork.ts
brendamanganelli: I don't think we need to differentiate them, as we are keeping everything pretty generic naming-wise!

src/Helpers/Click/ClickedEditArtwork.ts
brendamanganelli: Yes, it would be great to have it, I apologise for making it unclear - I'll amend the spreadsheet. Thanks for flagging this!

src/Helpers/Click/ClickedEditArtwork.ts
mdole: Sounds good!

src/Helpers/Click/ClickedEditArtwork.ts
mdole: No worries, thanks for clarifying!

mdole:
@damassi thanks for the help on this! Removed the helpers; I agree they're not necessary. I think it's ready to merge!

artsyit:
:rocket: PR was released in [`v1.79.0`](https://github.com/artsy/cohesion/releases/tag/v1.79.0) :rocket:

PR #173, opened 2021-03-19
Clarifies the use of helpers

sweir27:
The type of this PR is: **Documentation**

### Description

There has been some confusion recently around the role and importance of helper functions in this library. The goal of this PR is to clarify a little bit!

While some of the helper functions abstract away additional complexity, many end up only setting a few defaults or converting snake_case into camelCase. We've found in these cases it is easier (and less complex) to import the interfaces directly into the code to ensure that properties are properly typed, without passing them through a helper.


artsyit:
:rocket: PR was released in [`v1.78.0`](https://github.com/artsy/cohesion/releases/tag/v1.78.0) :rocket:

PR #175, opened 2021-03-22
Add artistAuctionResults owner

mikehrom:

This PR relates to [AS-1973].

### Description

We made the decision to keep web and app in sync with each other, so all views to the Auction Results (on web) or Insights (on app) tab will fire a screen view of type: artistAuctionResults, which should transform as "artist-auction-results". Changes to our pagetypes transforms will follow in subsequent fulcrum PRs, along with data migrations to use the new typings.

@MounirDhahri any existing instrumentation where the screen or page owner is "artistInsights" will need to be updated to "artistAuctionResults" - happy to pair on this when it can be prioritized!

[AS-1973]: https://artsyproduct.atlassian.net/browse/AS-1973

### DIFF
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -148,8 +148,8 @@ export interface TappedAuctionGroup extends TappedEntityGroup {
  *  ```
  *  {
  *    action: "tappedAuctionResultGroup",
- *    context_module: "auctionResults",
- *    context_screen_owner_type: "artistInsights",
+ *    context_module: "artistAuctionResults",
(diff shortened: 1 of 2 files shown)


MounirDhahri (approved):

MounirDhahri:
Follow up ticket for eigen here: https://artsyproduct.atlassian.net/browse/CX-1258

artsyit:
:rocket: PR was released in [`v1.80.0`](https://github.com/artsy/cohesion/releases/tag/v1.80.0) :rocket:

PR #178, opened 2021-04-01
Chore: Change babel config

mdole:
The type of this PR is: Bugfix

### Description

Slack context: https://artsy.slack.com/archives/CP9P4KR35/p1617223446228800

After adding Cohesion to Volt, we started having problems because one of Cohesion's exported JS files, ClickedShowMore.js, included a spread operator (`...`) and Volt wasn't able to easily transpile that feature.

We noticed that Cohesion was targeting `node` and figured we should update to target browsers instead, since it is consumed by web apps.

This PR updates the .babelrc targets to use Babel's `preset-env` default targets. We also added `corejs` as a dep since that's best practice according to [Babel docs](https://babeljs.io/docs/en/babel-preset-env). Thanks to @damassi for helping!

Here's a before/after for the offending file after compilation:

Before:
Screen Shot 2021-04-01 at 11 38 18 AM

After:
Screen Shot 2021-04-01 at 11 37 39 AM

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,9 @@
 {
   "presets": [
-    ["@babel/preset-env", { "targets": { "node": "current" } }],
+    [
+      "@babel/preset-env",
+      { "targets": "defaults", "useBuiltIns": "usage", "corejs": 3 }
+    ],
     "@babel/preset-typescript"
   ]
 }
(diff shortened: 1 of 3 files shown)


damassi (approved):
Yep, that before / after is exactly what I was expecting! Looks great 👍

artsyit:
:rocket: PR was released in [`v1.82.0`](https://github.com/artsy/cohesion/releases/tag/v1.82.0) :rocket:

PR #179, opened 2021-04-02
Adds Fair Items Event to CMS Checklist

brendamanganelli:
The event ClickedAddWorksToFair would fire when a partner with an upcoming booth clicks on Add Works button from CMS Checklist fair item. For reference, the flow is [here](https://www.figma.com/file/x19uDFCmGCfl2fbo9MKier/Untitled?node-id=0%3A1). Since this is really specific, I thought of adding an ad-hoc page owner type for the destination page. 

- Added ClickedAddWorksToFair event;
- Added new partnerShowsArtworks owner type

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -16,6 +16,33 @@ import { ActionType } from "."
  */
 
 /**
+ * A fair partner with an upcoming booth clicks on Add Works from CMS Checklist To-Do fair item.
+ *
+ *  This schema describes events sent to Segment from [[clickedAddWorksToFair]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "clickedAddWorksToFair",
+ *    context_module: "toDoList",
+ *    context_page_owner_type: "home",
+ *    destination_page_owner_type: "partner_shows",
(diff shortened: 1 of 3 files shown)


mdole (approved):
Looks great to me!

mikehrom:
we chatted about this in slack, but want to note here:
-- consider using `show` page owner type for the destination of this click, since the modal that opens is on an individual show page
-- we can leave `partnerShowArtworks` in page owner types, no harm done there

artsyit:
:rocket: PR was released in [`v1.83.0`](https://github.com/artsy/cohesion/releases/tag/v1.83.0) :rocket:

mdole:
Thanks for reviewing @mikehrom! I think on further reflection `partnerShowsArtworks` might make sense for the destination page owner type since this is specifically the artworks tab within an individual show (e.g. https://cms-staging.artsy.net/partner_shows/xavier-hufkens-xavier-hufkens-at-art-brussels-2021/artworks?current_partner_id=xavier-hufkens). So it's like...the artworks view within an individual show.

That said, I defer to you and @brendamanganelli! Let me know what you decide and I'll go with that

PR #180, opened 2021-04-06
chore: add label and remove type from ClickedEditArtwork

mdole:
The type of this PR is: enhancement

Paired with @brendamanganelli on reviewing checklist tracking, and we decided to add a `label` property to `ClickedEditArtwork` so that we could tell if the user clicked on a button to take them to the images flow or the price edit flow. We also removed `type` because it was unnecessary.

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -33,7 +33,7 @@ import { ActionType } from "."
  *  }
  * ```
  */
- export interface ClickedAddWorksToFair {
+export interface ClickedAddWorksToFair {
   action: ActionType.clickedAddWorksToFair
   context_module: ContextModule
   context_page_owner_type: PageOwnerType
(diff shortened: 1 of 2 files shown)


src/Schema/Events/Click.ts
```diff
@@ -33,7 +33,7 @@ import { ActionType } from "."
  *  }
  * ```
  */
- export interface ClickedAddWorksToFair {
+export interface ClickedAddWorksToFair {
```
mdole, line 36: I think this is just my auto-formatter having a good time 🤷

brendamanganelli:
Looks great, I'm merging it!

artsyit:
:rocket: PR was released in [`v1.84.0`](https://github.com/artsy/cohesion/releases/tag/v1.84.0) :rocket:

PR #184, opened 2021-04-26
refactor(filter params): replace nested properties with a JSON string

anandaroop:

This PR resolves [FX-2850]

### Description

Changes the shape of the generated `CommercialFilterParamsChanged` event so that it doesn't result in automatic creation of a zillion columns in Redshift. 

~We simply take some of the nested properties and replace them with a flat JSON string.~

Take 2! We expect Cohesion clients (currently that means only non-auction grids in Force) to send `current` and `changed` as JSON instead of as nested params. These will be forwarded unchanged to Segment/Redshift.

Basically this —

Screen Shot 2021-04-29 at 5 56 53 PM

~The public interface of the helper `commercialFilterParamsChanged` remains unchanged. Since this requires no change from upstream consumers (Force, Eigen) I am leaving this as a minor change re: semantic versioning, rather than a major/breaking change.~ 

This is now a breaking change for Force, so I'm inclined to bump the major version number.

It will also create breaking changes downstream, as @louislecluse as I have discussed.

[FX-2850]: https://artsyproduct.atlassian.net/browse/FX-2850

### DIFF
--- a/src/Helpers/FilterAndSort/CommercialFilterParamsChanged.ts
+++ b/src/Helpers/FilterAndSort/CommercialFilterParamsChanged.ts
@@ -1,14 +1,13 @@
 import {
   ActionType,
-  CommercialFilterParams,
   CommercialFilterParamsChanged,
   ContextModule,
   OwnerType,
 } from "../../Schema"
 
 export interface CommercialFilterParamsChangedArgs {
-  current: CommercialFilterParams
-  changed: CommercialFilterParams
+  current: string
+  changed: string
   contextOwnerId?: string
   contextOwnerSlug?: string
   contextOwnerType: OwnerType
@@ -23,8 +22,8 @@ export interface CommercialFilterParamsChangedArgs {
  *   context_owner_type: OwnerType.artist,
  *   context_owner_id: "58ba65b1275b24421f80a102",
  *   context_owner_slug: "tugo-cheng",
- *   changed: {acquireable: 1, priceRange: "1000-5000"},
- *   current: {acquireable: 0, priceRange: "*-*"}
+ *   changed: '{"acquireable": 1, "priceRange": "1000-5000"}',
+ *   current: '{"acquireable": 0, "priceRange": "*-*"}'
  * })
  * ```
  */
--- a/src/Helpers/FilterAndSort/__tests__/CommercialFilterParamsChanged.ts
+++ b/src/Helpers/FilterAndSort/__tests__/CommercialFilterParamsChanged.ts
@@ -4,27 +4,21 @@ import { commercialFilterParamsChanged } from "../CommercialFilterParamsChanged"
 describe("commercialFilterParamsChanged", () => {
   it("returns expected event properties", () => {
     const event = commercialFilterParamsChanged({
-      changed: { acquireable: true, priceRange: "1000-5000" },
+      changed: '{"acquireable":true,"priceRange":"1000-5000"}',
       contextOwnerId: "58ba65b1275b24421f80a102",
       contextOwnerSlug: "tugo-cheng",
       contextOwnerType: OwnerType.artist,
-      current: { acquireable: false, priceRange: "*-*" },
+      current: '{"acquireable":false,"priceRange":"*-*"}',
     })
 
     expect(event).toEqual({
       action: "commercialFilterParamsChanged",
-      changed: {
-        acquireable: true,
-        priceRange: "1000-5000",
-      },
+      changed: '{"acquireable":true,"priceRange":"1000-5000"}',
       context_module: "artworkGrid",
       context_owner_id: "58ba65b1275b24421f80a102",
       context_owner_slug: "tugo-cheng",
       context_owner_type: "artist",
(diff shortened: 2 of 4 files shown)


louislecluse:
Amazing thanks @anandaroop !
@mikehrom This might be of interest to you, we are changing the `commercial_filter_params_changed` event to only have a `changed` string containing all the filter entries. This simplifies the downstream redshift table and avoid to have a `zillion` column! Might be good to also do this for the auction results filters.

mikehrom:
This would be very nice to do with `auction_result_filter_params_changed` as well - thanks for giving me a heads up!

olerichter00 (commented):
Looks good so far!

Changing the helper will only affect **Force** though. In **Eigen** we track the event without using the helper (https://github.com/artsy/eigen/blob/09650eab62b71b7c773ab1737529318a8bd638c5/src/lib/Components/ArtworkFilter/ArtworkFilter.tsx#L109). 

Note: Changing the shape of the event in **Eigen** will affect the event `auctionResultsFilterParamsChanged` as well (https://github.com/artsy/eigen/blob/09650eab62b71b7c773ab1737529318a8bd638c5/src/lib/Components/ArtworkFilter/ArtworkFilter.tsx#L236)

anandaroop:
Good catch, @olerichter00! I will look into those.

anandaroop:
Okay 😳 

I've been staring at this for a while trying to figure out what strands I can pull on incrementally, and it's _quite_ complicated due to the divergent way in which Force and Eigen handle these filter changes, at least with respect to TS typing…

Force has strict typing around the shape of the filter params, and uses Cohesion to build a tracking event:

|Force types|Force event|
|---|---|
|Force|Force tracking|

Eigen has deliberately relaxed the typing around filter params, and does _not_ use Cohesion to build an event:

|Eigen types|Eigen event|
|---|---|
|Eigen|Eigen tracking|


And that's **before** we even consider the case of Auction grids, which are built in a very different way than all the other artwork grids.

So. In order to keep this moving here's what I'll do — going to **change my approach** so that 

1. I will update the `commercialFilterParamsChanged` here in Cohesion to accept a JSON `string` for `current` and `changed` params. (This will make it easier to convert Eigen and its looser typing to work with Cohesion in the future)

2. I will update Force to use this newer version of Cohesion send the json string to `commercialFilterParamsChanged`

3. Deploy that version of Force. 

4. I will update the non-Cohesion events being sent from Eigen to send JSON strings as well

5. Get that into an upcoming release of Eigen

At that point all filter param events (other than old versions of Eigen in the wild 😬) will have the new shape, and @louislecluse can do whatever is necessary on the Redshift side.

I will also ticket these as follow-ups but will not include them in the scope here, because my reading of the code suggests these will be pretty meaty work items in their own right:

- Convert Eigen's filter param events to use Cohesion
- Convert auction grids to use JSON instead of nested params

louislecluse:
Thanks @anandaroop 
Good to know about old eigen versions and auctions grid. That means we will keep having the legacy format flowing in, good to know. If we can fix auction grids as well, that would be best, it will be confusing in the future to have different types of events firing.
For auctions, is it the case both and web and iOS?

With that new plan, will `auction_result_filter_params_changed` still be impacted?

anandaroop:
> With that new plan, will auction_result_filter_params_changed still be impacted?

Not as part of FX-2850, but I will add a separate ticket for that.

louislecluse:
Sorry I think I wasn't clear on `auction_result_filter_params_changed`.
- Will this PR impact `auction_result_filter_params_changed`?
- If no, I will let @mikehrom decide if he wants to use this pattern now. If yes, can we avoid it?

anandaroop:
Ok, after some [more discussion](https://artsy.slack.com/archives/C9SATFLUU/p1619729438043600) to clarify that^ — 

- _This_ PR won't affect auction params anywhere
- The next PR in Force can include an update to JSON-ify auction filter params from web
- The next PR in Eigen will also JSON-ify auction filter params from iOS

anandaroop:
@louislecluse if you could just confirm that we are aligned on that^ re: auction filter events

mikehrom:
I will probably tag in when those subsequent PRs come up, especially the one involving Eigen.

louislecluse:
Sounds good @anandaroop . @mikehrom If you are using the filtered values for `auction_result_filter_params_changed` (what is actually clicked on in the filter), that logic will have to change with the update.

@anandaroop Are you also going to ticket the JSON-ification of the auction **grids**?

anandaroop:
> @anandaroop Are you also going to ticket the JSON-ification of the auction grids?

I will include that in the upcoming Force and Eigen PRs respectively. 

Might be a good idea to add these as subtasks for the current ticket though, so they can be tracked for visibility by all. I'll do that.

anandaroop:
Ok merging now, so that I can follow up in Force.

artsyit:
:rocket: PR was released in [`v2.0.0`](https://github.com/artsy/cohesion/releases/tag/v2.0.0) :rocket:

PR #185, opened 2021-05-05
Remove track api call for `screen` 

mikehrom:
The type of this PR is: **Bugfix, sorta**

This PR relates to [AS-2136]

### Description

I introduced a Track api call event named `screen` to the cohesion schema, but we really want a Screen api call.
Screen Shot 2021-05-05 at 11 13 29 AM

The pattern we want is how the Artist screen appears in the screenshot above, but right now for a few flows (myCollection for example), we fire a Track with the name `screen`. The former appears in `eigen_production.screens` (desired), the latter is in `eigen_production.screen`. We use the former in our sessions rollups, not the latter.

Want to get this right for Android launch, so we're starting with the first step of removing the track we do not want to use, and then we'll add structure in cohesion to differentiate between Screen api calls and Track api calls. For now, the non-cohesion schema for screens is acceptable (and preferred!).

[AS-2136]: https://artsyproduct.atlassian.net/browse/AS-2136

### DIFF
--- a/src/Schema/Events/System.ts
+++ b/src/Schema/Events/System.ts
@@ -55,27 +55,3 @@ export interface SaleScreenLoadComplete {
   load_time_ms: number
   number_of_lots: number
 }
-
-/**
- * A user views a screen on iOS
- *
- * This schema describes events sent to Segment from [[screen]]
- *
- *  @example
- *  ```
- *  {
- *    action: "screen",
- *    context_screen_owner_type: "artist",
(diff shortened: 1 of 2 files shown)


pvinis (approved):

abhitip:
LTGM - Is this going to break implementation? @pvinis will you follow up with removing those events?

pvinis:
I'll go ahead and close this actually. We do use this in eigen 😬. I thought we didn't.

PR #195, opened 2021-06-11
Saved Search 1.0 tracking

louislecluse:
The type of this PR is: **New events**

This PR resolves [FX-3014](https://artsyproduct.atlassian.net/browse/FX-3014)

### Description

Saved Search 1.0 tracking!
There is 1 event to track: which is toggling on/off the Saved Search! This event depends on how we model Saved Search on the back end and if we are able to generate a "saved search id" - I started the convo with engineering.

@abhitip I would like to instrument the push notification we send for a saved search alert. Is there anything to be done here?

- [ ] I've added comments with examples for any new interfaces or helpers and ensured that they're in the docs 

### DIFF
--- a/src/Schema/Events/Toggle.ts
+++ b/src/Schema/Events/Toggle.ts
@@ -59,7 +59,7 @@ export interface ToggledSavedSearch {
   context_screen_owner_type: ScreenOwnerType
   context_screen_owner_id?: string
   context_screen_owner_slug?: string
-  modified: number
-  original: number
+  modified: boolean
+  original: boolean
   search_criteria_id: string
 }
--- a/src/Schema/Events/Toggle.ts
+++ b/src/Schema/Events/Toggle.ts
@@ -47,6 +47,7 @@ export interface ToggledNotification {
  *    action: "toggledSavedSearch",
  *    context_screen_owner_type: "artist",
  *    context_screen_owner_id: "58de681f275b2464fcdde097",
+ *    context_screen_owner_slug: "anthony-hunter",
  *    modified: 1,
  *    original: 0,
  *    saved_search_id: "58de681f275b2464fcdde097"
@@ -57,6 +58,7 @@ export interface ToggledSavedSearch {
   action: ActionType.toggledSavedSearch
   context_screen_owner_type: ScreenOwnerType
   context_screen_owner_id?: string
+  context_screen_owner_slug?: string
   modified: number
   original: number
   saved_search_id: string
--- a/src/Schema/Events/Toggle.ts
+++ b/src/Schema/Events/Toggle.ts
@@ -48,8 +48,8 @@ export interface ToggledNotification {
  *    context_screen_owner_type: "artist",
  *    context_screen_owner_id: "58de681f275b2464fcdde097",
  *    context_screen_owner_slug: "anthony-hunter",
- *    modified: 1,
- *    original: 0,
+ *    modified: True,
+ *    original: False,
  *    search_criteria_id: "58de681f275b2464fcdde097"
  *  }
  * ```
--- a/src/Schema/Events/Toggle.ts
+++ b/src/Schema/Events/Toggle.ts
@@ -35,3 +35,29 @@ export interface ToggledNotification {
   original: number
   subject: PushNotificationType
 }
+
+/**
+ * A user toggles a saved search on/off on iOS
+ *
+ *  This schema describes events sent to Segment from [[toggledNotification]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "toggledSavedSearch",
+ *    context_screen_owner_type: "artist",
+ *    context_screen_owner_id: "58de681f275b2464fcdde097",
+ *    modified: 1,
+ *    original: 0,
+ *    saved_search_id: "58de681f275b2464fcdde097"
+ *  }
+ * ```
+ */
+export interface ToggledSavedSearch {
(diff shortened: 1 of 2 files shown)


src/Schema/Events/Toggle.ts
```diff
@@ -35,3 +35,29 @@ export interface ToggledNotification {
...
+ *    modified: 1,
+ *    original: 0,
+ *    saved_search_id: "58de681f275b2464fcdde097"
+ *  }
+ * ```
+ */
+export interface ToggledSavedSearch {
+  action: ActionType.toggledSavedSearch
+  context_screen_owner_type: ScreenOwnerType
+  context_screen_owner_id?: string
```
abhitip, line 59: what about context_screen_owner_slug?

abhitip:
Hmmm this push notification would be a product push notif right? Have you asked engineering?

src/Schema/Events/Toggle.ts
louislecluse: added!

src/Schema/Events/Toggle.ts
```diff
@@ -35,3 +35,31 @@ export interface ToggledNotification {
...
+  context_screen_owner_slug?: string
+  modified: number
+  original: number
+  saved_search_id: string
```
dblandin, line 64: We have two models in the database `SearchCriteria` and `UserSearchCriteria`. The `SearchCriteria` model holds the interesting bits, including the criteria attributes. `UserSearchCriteria` is the "join" model which exists to associate a user with a particular search criteria.

In the app, we have access to the `search_criteria_id`. If that's what we want to track here, what do you think of updating this attribute name to match?

src/Schema/Events/Toggle.ts
```diff
@@ -35,3 +35,31 @@ export interface ToggledNotification {
...
+ *    context_screen_owner_type: "artist",
+ *    context_screen_owner_id: "58de681f275b2464fcdde097",
+ *    context_screen_owner_slug: "anthony-hunter",
+ *    modified: 1,
+ *    original: 0,
+ *    saved_search_id: "58de681f275b2464fcdde097"
+ *  }
+ * ```
+ */
+export interface ToggledSavedSearch {
```
dblandin, line 57: Do we usually model a single `Toggle` event to capture two different states or would distinct `EnabledSavedSearch` and `DisabledSavedSearch` be useful?

src/Schema/Events/Toggle.ts
```diff
@@ -35,3 +35,31 @@ export interface ToggledNotification {
...
+  action: ActionType.toggledSavedSearch
+  context_screen_owner_type: ScreenOwnerType
+  context_screen_owner_id?: string
+  context_screen_owner_slug?: string
+  modified: number
+  original: number
```
dblandin, line 63: If we expect a boolean state, either "enabled" or "disabled", do our analytics support booleans or is the numeric 0/1 representation more compatible here?

src/Schema/Events/Toggle.ts
louislecluse: I think this goes back to [this](https://artsyproduct.atlassian.net/browse/FX-2808): the definition of the Saved Search model. From an analytics and domain model data standpoint we would need 1 object which holds the user saved search info and we would use the identifier here. I think we should resolve FX-2808 before proceeding here.

src/Schema/Events/Toggle.ts
louislecluse: I think toggle is ok! We can use the modified/original properties to figure out what happened!

src/Schema/Events/Toggle.ts
louislecluse: We can support booleans too. Do you suggest to update to "enabled": True/False, instead of modified/original?

dblandin: Sure, that would be great 👍🏼

src/Schema/Events/Toggle.ts
louislecluse: Going back to this after our [slack conversation](https://artsy.slack.com/archives/C9SATFLUU/p1623862368096700). Ideally we would use the  `UserSearchCriteria` identifier for `saved_search_id`. I want to directly identify the saved search user toggled. 
However, given that it seems we can only access the `search_criteria_id` in the app, and given all users are logged in in the app, then we should be able to retrieve the `UserSearchCriteria` object by using the `search_criteria_id` and the `user_id`. So seems it is fine to use the `search_criteria_id`!
Thank you @dblandin

src/Schema/Events/Toggle.ts
```diff
@@ -35,3 +35,31 @@ export interface ToggledNotification {
...
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "toggledSavedSearch",
+ *    context_screen_owner_type: "artist",
+ *    context_screen_owner_id: "58de681f275b2464fcdde097",
+ *    context_screen_owner_slug: "anthony-hunter",
+ *    modified: True,
+ *    original: False,
```
dblandin, line 52: **minor**: in JS, boolean values are true / false (lowercase)

dblandin (approved):
LGTM! Thanks for the updates 👍

abhitip:
LGTM! I'll let you merge

artsyit:
:rocket: PR was released in [`v2.4.0`](https://github.com/artsy/cohesion/releases/tag/v2.4.0) :rocket:

PR #196, opened 2021-06-15
[PX-4111] - Adds new shipping flow events to Cohesion

brendamanganelli:
1. I have added new events to being able to track users' interactions with the new shipping flow for **web**; designs are [here](https://www.figma.com/file/J5bIQj4a0l7wYAHnhbVGTp/%5BCollector%5D-Checkout-Shipping?node-id=1761%3A0);
2. Added two new context_module: OrdersShipping and OrdersReview to distinguish between orders-shipping and orders-review pagetypes.

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -494,7 +494,6 @@ export interface ClickedNavigationTab {
   context_page_owner_id: string
 }
 
-
 /**
  * A user clicks a partner card
  *
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -16,6 +16,29 @@ import { ActionType } from "."
  */
 
 /**
+ *  User clicks to add new shipping address when entering the orders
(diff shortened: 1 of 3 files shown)


mikehrom (commented):
I have general questions -- other than clicking the buyer protection link (which might indicate people are interested in what that means and what they gain from the feature), what do we hope to learn with all these other clicks? Shipping method and pickup/ship are stored on the order... are we goaling against any interactions for this feature? I don't want the tracking to be too heavy-handed is all.

brendamanganelli:
Hi! The idea was to have a clearer picture on where users lapse / abandon the checkout flow. I agree not to be so granular though - and maybe tracking pageviews is enough when it comes to orders! I would love to keep the event related to the buyer protection link and the events associated to the Orders Review page. What do you think?

artsyit:
:rocket: PR was released in [`v2.10.0`](https://github.com/artsy/cohesion/releases/tag/v2.10.0) :rocket:

PR #203, opened 2021-07-06
[PX-4220] - Track analytics for weight & framed dimensions in CMS

brendamanganelli:
- Added new context module for Volt Artwork Edit page
- Added events for weight and & framed dimensions in CMS


src/Schema/Events/index.ts
```diff
@@ -22,6 +22,9 @@ import {
   ClickedLoadMore,
   ClickedMainArtworkGrid,
   ClickedNavigationTab,
+  ClickedOnFrameMeasurements,
```
pepopowitz, line 25: The framed measurements portion of the form consists of three input fields. Does "clicked on" capture exactly what we want to track? Do we care which field they focused? Does tabbing to a field mean the same thing as clicking on it? Do we care if they entered values?

src/Schema/Events/index.ts
```diff
@@ -22,6 +22,9 @@ import {
...
+  ClickedOnArtworkShippingUnitsDropdown,
```
pepopowitz, line 26: Does "clicked on" mean "focused", since they could tab to a dropdown? Do we care if they select a field or just that they focused on the field?

pepopowitz (commented):
Thank you!

I have questions about what these events mean in the context of filling out the artwork form, since entering input is different than clicking on a button in force.

src/Schema/Events/index.ts
pepopowitz: Thanks Brenda! 

It sounds like we want to fire an event when they focus on each field (height, width, depth, or diameter), so that it fires when they click or tab into the field. I don't see any mechanism in this method to distinguish between _which_ field they focused on -- is that okay?

artsyit:
:rocket: PR was released in [`v2.12.0`](https://github.com/artsy/cohesion/releases/tag/v2.12.0) :rocket:

pepopowitz:
I merged this as-is so we can move forward with implementation, but @brendamanganelli when you have some time can you take a look at https://github.com/artsy/cohesion/pull/203#discussion_r672378286 to see if everything is covered with these changes?

src/Schema/Events/index.ts
brendamanganelli: Thank you! Yes, this is ok! We want to avoid such level of granularity in Cohesion; we can still see bottleneck areas or sense patterns by analysing these interactions together with the data we can get from redshift in regards to these fields.

PR #214, opened 2021-08-03
Schema for Price Database web events

mikehrom:
Schema for events on the new Price Database experience for web. Wireframes [here](https://www.figma.com/file/4fxnvYKVZzBr3TEE6WBj2J/Price-Database-%2F-Q3-2021?node-id=1226%3A8174) and comments in-line.

### DIFF
--- a/src/Schema/Events/Search.ts
+++ b/src/Schema/Events/Search.ts
@@ -138,3 +138,31 @@ export interface FocusedOnPriceDatabaseSearchInput {
   context_module: ContextModule.priceDatabaseLanding
   context_owner_type: OwnerType.priceDatabase
 }
+
+/**
+ * A user selects an artist from the Price Database artist search
+ *
+ * This schema describes events sent to Segment from [[selectedItemFromPriceDatabaseSearch]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "selectedItemFromPriceDatabaseSearch",
+ *    context_module: "priceDatabaseLanding",
+ *    context_owner_type: "priceDatabase",
+ *    owner_type: "artist",
+ *    owner_id: "5df3e3fa485efe0012c37055",
+ *    owner_slug: "andy-warhol",
+ *    query: "andy warhol"
+ *  }
+ * ```
+ */
+export interface SelectedItemFromPriceDatabaseSearch {
+  action: ActionType.selectedItemFromPriceDatabaseSearch
+  context_module: ContextModule.priceDatabaseLanding
+  context_owner_type: OwnerType.priceDatabase
+  owner_type: OwnerType.artist
+  owner_id: string
+  owner_slug: string
+  query: string
+}
--- a/src/Schema/Events/Search.ts
+++ b/src/Schema/Events/Search.ts
@@ -87,3 +87,52 @@ export interface SearchedWithNoResults {
   destination_owner_type?: PageOwnerType
   query: string
 }
+
+/**
+ * A user queries the Artsy Price Database
+ *
+ * This schema describes events sent to Segment from [[searchedPriceDatabase]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "searchedPriceDatabase",
+ *    context_module: "priceDatabaseLanding",
+ *    context_owner_type: "priceDatabase",
+ *    destination_owner_type: "artistAuctionResults",
+ *    destination_owner_id: "4e9746234e77d300010010c6",
+ *    destination_owner_slug: "franz-kline",
+ *    query: "franz kline"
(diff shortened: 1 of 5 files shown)


src/Schema/Events/Search.ts
```diff
@@ -87,3 +87,52 @@ export interface SearchedWithNoResults {
...
+ * ```
+ */
+export interface SearchedPriceDatabase {
+  action: ActionType.searchedPriceDatabase
+  context_module: ContextModule.priceDatabaseLanding
+  context_owner_type: OwnerType.priceDatabase
+  destination_owner_type: OwnerType.artistAuctionResults
+  destination_owner_id: string
+  destination_owner_slug: string
+  query: string
```
mikehrom, line 117: Something to determine: do we pass only the string from the search box or do we include search params? The search params will be on the filtering events (`current`, `changed` props) but I believe we also intend to land users on the artist-insights tab with filters applied, based on their search from the price-database landing page.

src/Schema/Values/ContextModule.ts
```diff
@@ -96,6 +96,7 @@ export enum ContextModule {
   popUpModal = "popUpModal",
   presentingFair = "presentingFair",
   presentingPartner = "presentingPartner",
+  priceDatabaseLanding = "priceDatabaseLanding",
```
mikehrom, line 99: feedback welcome on the context module name, but I thought this summed up the general idea for where users will be searching

src/Schema/Events/Search.ts
```diff
@@ -87,3 +87,52 @@ export interface SearchedWithNoResults {
...
+ *    context_owner_type: "priceDatabase",
+ *    destination_owner_type: "artistAuctionResults",
+ *    destination_owner_id: "4e9746234e77d300010010c6",
+ *    destination_owner_slug: "franz-kline",
+ *    query: "franz kline"
+ *
+ *  }
+ * ```
+ */
+export interface SearchedPriceDatabase {
```
mikehrom, line 110: The gist of this is that users will be searching artist names in combination with some filter params (size, time period, medium) on a specialized landing page/entry point to the Price Database. My understanding is that artists are always sent to the artist-insights tab on the Artist page upon entering a query, so the context and destination props reflect that.

src/Schema/Events/Search.ts
louislecluse: Looks fine to me to separate filter and search events and properties.
The advantage of having filter and search in the same event is that it will be easier to find the whole query (artist + filters) the user did. For that you could instrument the search button (clicked search button event) and pass both the query and the filters are properties.
But having just filter and search as separate events might more natural and clearer for now!

Q: Does the user has to select an artist similar to our autosuggest experience in the search bar? That might be a simpler event to track and will give more structured data (such as artist id) than the queries which are just text.

louislecluse (approved):

src/Schema/Events/Search.ts
olerichter00: @louislecluse The autosuggest input works exactly as the "Search by artist name" input on https://www.artsy.net/consign

I think we can either include the query string (where we pass the filters) or the filters (currently applied filters when the user clicks "search") in this event.

olerichter00 (approved):

src/Schema/Events/Search.ts
mikehrom: the fact that this feature works like 'search by artist name' on /consign makes me want to re-do these events to match, especially if we can pass `query` and `filters` when a search is made.

mikehrom: > Does the user has to select an artist similar to our autosuggest experience in the search bar? That might be a simpler event to track and will give more structured data (such as artist id) than the queries which are just text.

@louislecluse i think i realize what's bothering me about possibly using 'selectedItemFromSearch' with a context module for priceDatabase, and it's that within `session_tracks` we roll up all 'selectedItemFromSearch' tracks with no regard for context into one measure per session. if we used that event here, we'd be mixing contexts (which are pretty different user flows, imo), and what we're currently meaning to aggregate is the selection of artists from search in the nav bar. 

i think all things considered, i'll do this:
- continue with a new event (searchedPriceDatabase) with properties for the query _and_ the filters passed
- defer to CX about implementing tracks on the filters themselves, since we'll be getting selected filters in the aforementioned track
- defer to CX about implementing focusedOnPriceDatabaseSearchInput, since again what we really need is searchedPriceDatabase, the others are secondarily important I think

artsyit:
:rocket: PR was released in [`v2.16.0`](https://github.com/artsy/cohesion/releases/tag/v2.16.0) :rocket:

PR #216, opened 2021-08-10
Event for failed artist search in the consignments flow

mikehrom:
Consignments business team wants to know who users are attempting to submit artworks for outside of our P1/P2 list; this event will provide a way to query for those search bar inputs. Example of where this happens on web:

Screen Shot 2021-08-10 at 2 22 26 PM


artsyit:
:rocket: PR was released in [`v2.17.0`](https://github.com/artsy/cohesion/releases/tag/v2.17.0) :rocket:

PR #217, opened 2021-08-11
Enhancement(CX-1698): add link taps action and screen auctionResultsForArtistYouFollow to cohesion

lordkiz:

This PR resolves [CX-1800]

### Description
- Adds a general ActionType for taps on inline Hypertext/links

- Adds OwnerType.auctionResultsForYou to ScreenOwnerTyper to allow for tracking of ActionResultsForYou screen

[CX-1800]: https://artsyproduct.atlassian.net/browse/CX-1800


src/Schema/Events/Tap.ts
```diff
@@ -647,3 +647,31 @@ export interface TappedVerifyIdentity {
...
+ */
+export interface TappedLink {
+  action: ActionType.tappedLink
+  context_module: ContextModule
+  context_screen_owner_type: ScreenOwnerType
+  context_screen_owner_id?: string
+  context_screen_owner_slug?: string
+  destination_screen_owner_type?: ScreenOwnerType
+  destination_screen_owner_id?: string
+  destination_screen_owner_slug: string
```
mikehrom, line 675: i think maybe `destination_screen_owner_slug` is optional too, unless you're referring to `destination_path` -- which I believe we use as a property in other events. `destination_path` works for internal or external links, where external links won't have a `slug` that corresponds to some artsy data object (like an artwork, artist, or partner).

mikehrom: tl;dr:
- suggesting `destination_path` as required prop
- suggesting `destination_screen_owner_slug?`

mikehrom (commented):
thanks for this! i commented in line with one suggestion. we can sync on slack or zoom if you need, too.

src/Schema/Events/Tap.ts
mikehrom: Correct, I'd agree with that 👍

lordkiz: Ok thanks

artsyit:
:rocket: PR was released in [`v2.19.1`](https://github.com/artsy/cohesion/releases/tag/v2.19.1) :rocket:

PR #220, opened 2021-08-13
Saved Search M2 tracking

louislecluse:
Tracking for Saved Search M2.


dblandin:
@louislecluse Do these events describe the initial tap or when the action is performed?

For example, a user can click "Create alert" but abandon the flow. Similarly for deleting a saved search: a user can click "Remove alert" but then decide to keep it and dismiss the confirmation prompt.

louislecluse:
For edit and delete, I was thinking of tracking when the action is performed.

For create alert I was thinking when user taps the "create alert" button, we already have the event that is fired when the alert is created ("toggled saved search").

dblandin:
> For edit and delete, I was thinking of tracking when the action is performed.

In that case, should we rename the events to reflect that it's not just a tap but after the action is performed? Something like `editedSavedSearch` and `deletedSavedSearch`?

louislecluse:
Good point, added relevant events.

dblandin (approved):

artsyit:
:rocket: PR was released in [`v2.19.0`](https://github.com/artsy/cohesion/releases/tag/v2.19.0) :rocket:

Issue #212, opened 2021-08-02
RFC: Remove helpers


pvinis:
The main suggestion here is to remove helpers from cohesion and pull them in the clients if they want to use them.

The usefulness of cohesion is the fact that it provides types and interfaces to make sure we send the right data to our analytics. How do that per client shouldn't matter to cohesion. It also makes PRs harder to create, as we need to have an engineer help make these helpers, before we even have the chance to use them, so the final look of a helper is still uncertain.

By removing the helpers from cohesion, we can allow for it to be owned by data team way more cleanly. It also allows the clients to decide how they want to handle sending. The types/interfaces are what data team needs, and what the clients can obey.

## Reasoning

It seems like the helpers of cohesion are a cause of duplication and confusion. There have been a few times already where helpers are not used and are duplicated in the clients, or times were there are no helpers created.

- `share`
The [share helper here](https://github.com/artsy/cohesion/blob/master/src/Helpers/Share.ts) was created, and in the end it's [not used](https://github.com/artsy/eigen/blob/master/src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx#L188-L203).

- `tappedShowMore`
[See here](https://github.com/artsy/cohesion/search?q=tappedShowMore) that cohesion has no helper, and eigen just uses [its own here](https://github.com/artsy/eigen/blob/7ae0c7e8aaa37ad78b4e3ffe9388cc3a15740e27/src/lib/Scenes/MyCollection/Screens/Artwork/Components/MyCollectionArtworkMeta.tsx#L102-L114).

There are more examples of this, I quickly found these for now.

For the first one, I remember adding this while being unsure how we will use this, and turns out it's easier to have two helpers, one for the native ios share, one for the new default, custom share. Cohesion doesn't need to know that. It's enough if both eigen helpers send the right things needed, based on the schema/types/interfaces.

The second one is one of the many examples where an engineer either was not available, or was unsure, or just didn't need to add a helper at that point. It works well.

As an extra, eigen already has an established way to do tracking [documented here](https://github.com/artsy/eigen/blob/7ae0c7e8aaa37ad78b4e3ffe9388cc3a15740e27/docs/analytics_and_tracking.md) using local helpers that can either live in the component file, or in a helper file if they are more general.

Edit: As @sweir27 accurately pointed out in a comment below, these helpers are not actually shared. force and eigen have their own. So it makes helpers even less neccessary to live within cohesion.

## Exceptions:

[...]

PR #233, opened 2021-09-22
Deprecate Helpers

pvinis:

### Description

As part of the accepted RFC https://github.com/artsy/cohesion/issues/212, I am making this PR to deprecate helpers. The next steps are to slowly remove them while we make changes to eigen and force to stop using them, bit by bit.


mdole (approved):
👍 nice, this is a good first step towards deprecation! In case others haven't seen it, this message has a bit more details on the overall path to fully removing these from Cohesion: https://artsy.slack.com/archives/C015KTU8RQT/p1632308619004300

pvinis:
I'll wait for at least one more review, then I will merge it.

damassi (approved):
Appreciatetd that these files were moved -- deprecated -- rather than deleted 👍 

LGTM! (Though we should probably add a `canary` tag to the PR as well and test things against that before merging.)

sweir27 (approved):
Agree with Chris about testing with a canary, but this is awesome! Thank you for continuing on with this work!!

pvinis:
I did make a canary on https://github.com/artsy/cohesion/pull/235 and https://github.com/artsy/cohesion/pull/236 and they both worked fine on eigen and force 👍.

Thank you all for the reviews. I'll go ahead and merge, and go from there for the example PRs.

artsyit:
:rocket: PR was released in [`v4.0.0`](https://github.com/artsy/cohesion/releases/tag/v4.0.0) :rocket:

dzucconi:
These should be marked with `@deprecated` so that our tooling picks it up?

dzucconi:
Also what's the migration path for tracking authentication events or do we just migrate the helpers over? They need to be outside of components so we won't be using the hooks.

dzucconi:
OK — I completely misunderstood what these functions did! I thought they were actually making calls to track!

pvinis:
I didn't think of the @deprecated comment. that would be cool, but maybe too much work? 🤔

PR #238, opened 2021-09-24
Sent inquiry

anipetrov:
The type of this PR is an enhancement of an existing event, `sent_artwork_inquiry`
 -->


src/Schema/Events/Inquiries.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/**
+ * The type of product associated with an inquiry request
+ */
+export type InquiryProduct = {
+  price: number
+  product_id: string
+  quantity: 1
+}[]
```
dzucconi, line 46: You'd type this without the array; then include the array form in the `products` field above (`products: InquiryProduct[]`)

anipetrov:
gah looks like i was on an old branch, will edit on monday

src/Schema/Events/Inquiries.ts
anipetrov: thanks!

dzucconi:
Apparently the product data is for Criteo? https://github.com/artsy/force/pull/4041

(Any ideas for the `6000` default? https://github.com/artsy/force/blob/938ae6d7ac57c93141052dd05821d15850dbab63/src/desktop/components/inquiry_questionnaire/analytics/events.ts#L168)

src/Schema/Events/Inquiries.ts
```diff
@@ -0,0 +1,46 @@
...
+ *  @example
+ *  ```
+ *  {
+ *    action: "sentArtworkInquiry",
+ *    artwork_id: "582b1bc09c18db3b0c000999",
+ *    artwork_slug: "philippe-halsman-alfred-hitchcock-during-the-filming-of-the-birds",
+ *    inquiry_id: "614e3063ef0b81000bc4f45a",
+ *    products: [
+        {
+          price: 6000,
```
dzucconi, line 22: I'm pretty sure this is a pre-formatted string most of the time? The default just happens to be 6000 for some reason?

abhitip:
Oh interesting, thanks for finding that @dzucconi ! I think we can remove the product id, right @anipetrov ? I feel like most marketing integrations I'm seeing these days will let us map the artwork id to their internal definition, rather than forcing us to name it how they'd like. Lmk what you think!

src/Schema/Events/Inquiries.ts
```diff
@@ -0,0 +1,46 @@
...
+  artwork_slug: string
+  inquiry_id: string
+  products: InquiryProduct
+}
+
+/**
+ * The type of product associated with an inquiry request
+ */
+export type InquiryProduct = {
+  price: number
```
dzucconi, line 43: This is likely a string — if we can check the data to find out what it's receiving that'd be great.

dzucconi:
I think `e305932` is causing this to fail in master. I'm going to revert then just rebase this.

PR #239, opened 2021-09-27
Fixes to types for new inquiry

dzucconi:
This PR corrects and adds a few things to support https://github.com/artsy/force/pull/8483

### DIFF
--- a/src/DeprecatedHelpers/Authentication/CreatedAccount.ts
+++ b/src/DeprecatedHelpers/Authentication/CreatedAccount.ts
@@ -34,6 +34,7 @@ export interface CreatedAccountArgs {
  *   userId: "5bd8b675776bd6002c86526c"
  * })
  * ```
+ * @deprecated
  */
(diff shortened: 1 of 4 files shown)


abhitip:
Why is auth_redirect optional now? Is it because it would just be the inquiry modal for inquiry?

dzucconi:
Yeah; it's not redirecting anywhere. You login and stay on the page and there's no redirect. Unless you want this param to be just whatever the current page is?

abhitip:
I think that's what we've done in the past - like if you login on the artist page, it just stays on the artist page right? I think auth_redirect would be the artist page in that case. Could we just keep it at that?

dzucconi:
@abhitip sure works for me. Updated.

artsyit:
:rocket: PR was released in [`v4.2.0`](https://github.com/artsy/cohesion/releases/tag/v4.2.0) :rocket:

PR #243, opened 2021-09-30
chore: move shipping constants into cohesion

pepopowitz:
The type of this PR is: Chore

### Description

Adds constants related to the new shipping flow. We hardcoded [a couple of them in force](https://github.com/artsy/force/blob/e1a53b84f651d63990510df21f6d9ec69486bab3/src/v2/Apps/Order/Routes/Review/index.tsx#L440-L441) to get things out the door; this PR moves them to the correct home. 

It also includes several constants not being used yet in force, but which will allow us to specify the context module/owner type for every instance the Buyer Guarantee is clicked. We're currently only tracking those clicks from the shipping & review pages of the order app, but the component lives on pretty much every page in the order app.


src/Schema/Values/OwnerType.ts
```diff
@@ -40,6 +40,14 @@ export enum OwnerType {
...
   myCollectionArtwork = "myCollectionArtwork",
   onboarding = "onboarding",
+  ordersAccept = "orders-accept",
+  ordersCounter = "orders-counter",
+  ordersNewPayment = "orders-new-payment",
+  ordersOffer = "orders-offer",
+  ordersPayment = "orders-payment",
+  ordersRespond = "orders-respond",
+  ordersReview = "orders-review",
+  ordersShipping = "orders-shipping",
```
pepopowitz, line 50: We are using these kebab-case values in the hardcoded force values, but everything else in this file is using camelCase. Is that a problem???

src/Schema/Values/ContextModule.ts
```diff
@@ -86,8 +86,14 @@ export enum ContextModule {
   marketInsights = "marketInsights",
   newWorksByArtistsYouFollowRail = "newWorksByArtistsYouFollowRail",
   newWorksByGalleriesYouFollowRail = "newWorksByGalleriesYouFollowRail",
-  ordersShipping = "ordersShipping",
```
damassi, line 89: Can alphabetize this? (noted that it was out of alpha to begin with)

pepopowitz: ugh, I thought I did but apparently I don't know the English alphabet 😅 

👍 Will do.

pepopowitz: Resolved in most recent commit!

pepopowitz:
@brendamanganelli could you provide a review for this so that I can move forward with it on the Force side?

src/Schema/Values/OwnerType.ts
brendamanganelli: This is great and it is consistent with what we've been doing!

brendamanganelli (approved):
No objections on my side! Ship it!

PR #248, opened 2021-10-07
refactor(FX-2978): remove authImpression deprecated helper

araujobarret:

This PR resolves [PURCHASE-2978](https://artsyproduct.atlassian.net/browse/PURCHASE-2978)

### Description

This PR removes the deprecated helper authImpression.

Should be merged only after [#8570](https://github.com/artsy/force/pull/8570) merge


mdole:
just double-checked to confirm this is no longer being used anywhere, looks like we're good to go! let's ship it

PR #255, opened 2021-10-13
Adding new events to Cohesion

audrey-thenot:

This PR resolves [CO-]


abhitip (approved):
looks good to me! You might just want to move the tappedContactGallery event to the Tap folder, rather than the Conversations folder.

PR #260, opened 2021-11-03
feat(CX-2146): add promptForReview event

MounirDhahri:

This PR resolves [CX-2146]

### Description

This PR adds `promptForReview` event. This event refers to asking the users to give an app review on both android and iOS.

[CX-2146]: https://artsyproduct.atlassian.net/browse/CX-2146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/PromptForReview.ts
+++ b/src/Schema/Events/PromptForReview.ts
@@ -28,5 +28,5 @@ export interface PromptForReview {
   context_owner_type: OwnerType
   context_owner_id?: string
   context_owner_slug?: string
-  subject: "ios" | "android"
+  platform: "ios" | "android"
 }
--- a/src/Schema/Events/PromptForReview.ts
+++ b/src/Schema/Events/PromptForReview.ts
@@ -1,0 +1,32 @@
+import { ContextModule } from "../Values/ContextModule"
+import { OwnerType } from "../Values/OwnerType"
+import { ActionType } from "."
+
+/**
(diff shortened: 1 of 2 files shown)


src/Schema/Events/PromptForReview.ts
```diff
@@ -0,0 +1,32 @@
...
+ *  }
+ * ```
+ */
+export interface PromptForReview {
+  action: ActionType.promptForReview
+  context_module: ContextModule
+  context_owner_type: OwnerType
+  context_owner_id?: string
+  context_owner_slug?: string
+  subject: "ios" | "android"
```
lordkiz, line 31: @MounirDhahri The example doesn't match. Do you mean platform?

MounirDhahri: good catch 😬

MounirDhahri: fixed

lordkiz (approved):
LGTM 🚀 . A follow up might be to track the prompt dismissal/review in order to determine the CTR.

artsyit:
:rocket: PR was released in [`v4.19.0`](https://github.com/artsy/cohesion/releases/tag/v4.19.0) :rocket:

MounirDhahri:
We thought about that but unfortunately, it is not possible as apple doesn't allow us to hook events to their native prompt

PR #291, opened 2022-01-05
chore: add createAlert to ContextModule

dimatretyak:

This PR resolves [CO-]

### Description
After I received [this answer](https://artsyproduct.atlassian.net/browse/FX-3545?focusedCommentId=44885), I found that there is no `createAlert` in the `ContextModule`. This PR fixes this problem

### DIFF
--- a/src/Schema/Values/ContextModule.ts
+++ b/src/Schema/Values/ContextModule.ts
@@ -48,6 +48,7 @@ export enum ContextModule {
   collectionDescription = "collectionDescription",
   collectionRail = "collectionRail",
   consignSubmissionFlow = "consignSubmissionFlow",
+  createAlert = "createAlert",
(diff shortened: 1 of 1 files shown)


src/Schema/Values/ContextModule.ts
```diff
@@ -7,6 +7,7 @@
 export enum ContextModule {
   aboutTheWork = "aboutTheWork",
   aboutThisAuction = "aboutThisAuction",
+  alert = "alert",
```
dimatretyak, line 10: @louislecluse maybe it's better to call `createAlert`?

dimatretyak: Renamed

artsyit:
:rocket: PR was released in [`v4.28.0`](https://github.com/artsy/cohesion/releases/tag/v4.28.0) :rocket:

PR #296, opened 2022-01-18
feat: add shareSheet context module

dimatretyak:

### Description
Add `shareSheet` context module for reusable `Share` component in Eigen


louislecluse:
Why do we need this context module?

dimatretyak:
> Why do we need this context module?

As part of the hackathon, a reused `Share` component **was created** and I planned to add a context module for it

PR #299, opened 2022-01-28
[AS-2900] Updates App submission flow events

abhitip:

This PR resolves [AS-2900]

- [ x] If I've added a new file to the tree I've exported it from the common `index.ts`
- [x ] I've added comments with examples for any new interfaces and ensured that they're in the docs
- [x ] No platform-specific terminology has been used outside of `click` and `tap` (platform is inferred by the DB storing events)

[AS-2900]: https://artsyproduct.atlassian.net/browse/AS-2900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Schema/Values/OwnerType.ts
```diff
@@ -22,6 +22,7 @@ export enum OwnerType {
   collect = "collect",
   collection = "collection",
   collections = "collections",
+  consignmentFlow = "consignmentFlow",
```
iskounen, line 25: **nitpick:** We're moving away from copy that says "consignment" on the frontend. Should we start using `sellWithArtsy...` instead of `consignment...`?

src/Schema/Values/ContextModule.ts
```diff
@@ -151,6 +153,7 @@ export enum ContextModule {
   trendingArtistsRail = "trendingArtistsRail",
   trendingLots = "trendingLots",
   upcomingAuctions = "upcomingAuctions",
+  uploadPhotosCompleted = "uploadPhotosCompleted",
```
iskounen, line 156: **suggestion:** Is it OK if these context modules get reused for tracking other events, like uploading photos for My Collection, or entering contact information on your user profile? If not, I think we should group them together so that it's clear that they belong to the Sell with Artsy submission flow.

src/Schema/Events/UserExperienceInteractions.ts
```diff
@@ -0,0 +1,32 @@
...
+import { OwnerType } from "../Values/OwnerType"
+import { ActionType } from "."
+
+/**
+ * Schemas describing events for specific user experience design types.
+ * @packageDocumentation
+ */
+
+/**
+ * A user expands or collapses an accordion design, like in the consignments flow.
```
iskounen, line 11: **suggestion:** Should there be a way to differentiate between an expand and collapse action?

src/Schema/Values/ContextModule.ts
abhitip: Yes it's okay if they get reused elsewhere!

src/Schema/Events/UserExperienceInteractions.ts
abhitip: I just added an `expand` boolean - thanks Adam!

src/Schema/Values/OwnerType.ts
abhitip: Good to know! I think I'll keep it as consignment because I think it is less wordy at describing the submission flow

iskounen (approved):

audrey-thenot:
Looks good to me!

audrey-thenot (approved):

PR #304, opened 2022-03-01
adding experiment

pvinis:

### Description

Paired with @mc-jones on this.

We used to have https://github.com/artsy/force/blob/main/src/v2/System/Analytics/trackExperimentViewed.tsx in force for Split experiments. Going to unleash for both eigen and force, we are adding a schema to cohesion for consistency.

The names of this event are different, lets us know if thats ok and if something is missing.

<details>
  <summary>📦 Published PR as canary version: <code>4.34.1--canary.304.4466.0</code></summary>
  <br />

  ```bash
  npm install @artsy/cohesion@4.34.1--canary.304.4466.0
  # or 
  yarn add @artsy/cohesion@4.34.1--canary.304.4466.0
  ```
</details>


src/Schema/Events/Experiment.ts
```diff
@@ -0,0 +1,46 @@
...
+  service: "unleash" | string
+  experiment_name: string
+  enabled: string
+
+  context_owner_type: OwnerType
+  context_owner_id?: string
+  context_owner_slug?: string
+} & (
+  | {
+      event: "get_flag"
```
abhitip, line 39: How does this part work? It's not a property is it?

abhitip (commented):
Thanks Pavlos! I'm going to run this by the data team tomorrow and then I'll let you know

src/Schema/Events/Experiment.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * This schema describes events sent to Segment from [[Experiment]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "experiment",
+ *    event: "get_variant",
+ *    service: "unleash",
+ *    experiment_name: "some_experiment_we_are_running",
```
pvinis, line 20:
```suggestion
 *    experiment_name: "some_experiment_we_are_running",
 *    enabled: true,
```

src/Schema/Events/Experiment.ts
pvinis: you mean the `&` and `|` part?

everything before `&` is what is needed, and then also everything after is needed. after `&` we have a paren with two options, marked with `|`, so one or the other is added.

so with that, we end up with a type that would allow us something like that:
```
{
    action: "experiment",
    event: "get_variant",
    service: "unleash",
    experiment_name: "some_experiment_we_are_running",
    variant_name: "some_variant_for_this_user"
    payload: "some_payload",
    context_owner_type: "artwork",
    context_owner_id: "55ed8ca57261693d930000b8",
}
```

it will allow:
```
{
    action: "experiment",
    event: "get_flag",
    service: "unleash",
    experiment_name: "some_experiment_we_are_running",
    context_owner_type: "artwork",
    context_owner_id: "55ed8ca57261693d930000b8",
}
```

but it will *not* allow this:
```
{
    action: "experiment",
    event: "get_variant",
    service: "unleash",
    experiment_name: "some_experiment_we_are_running",
// notice that `variant_name` is missing here!
    context_owner_type: "artwork",
    context_owner_id: "55ed8ca57261693d930000b8",
}
```

does that make sense?

src/Schema/Events/Experiment.ts
```diff
@@ -0,0 +1,47 @@
...
+ *  }
+ * ```
+ */
+export type Experiment = {
+  action: ActionType.experiment
+  service: "unleash" | string
+  experiment_name: string
+  enabled: string
+
+  context_owner_type: OwnerType
```
mc-jones, line 35: @pvinis - This should be optional, too, no?

pvinis: hm I remember @abhitip saying this shouldn't but I might be remembering wrong. should we make this optional?

abhitip (commented):
@pvinis Here's what we discussed!

- Change name to `experimentViewed`
- Remove `enabled` property because we only want this event to fire when test is live (which means enabled is True) for both control and experiment.
- Can you add this as a comment next to `variant_name` to remind us? "variant_name should be `control` or `experiment` when there are only 2 groups, and something more specific when there are > 2 groups (i.e. `control`, `big_header`, `small_header`)"
- Keep the payload and add a comment about how it's another input that can change a part of the feature for a user so this will be helpful to know for analysis.

Lmk if you have questions!

pvinis:
did some updates. should be good? 🤔

pvinis:
lol i put the `canary` label to check the docs, but i guess this is for just a canary release, not docs.

src/Schema/Events/ExperimentViewed.ts
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Schemas describing experiment events
+ * @packageDocumentation
+ */
+
+/**
+ * A user received an experiment on Artsy.
+ *
+ * This schema describes events sent to Segment from [[Experiment]].
```
abhitip, line 12: update this to experimentViewed?

abhitip:
Just had some minor comments to update the documentation, but otherwise looks great! Thanks @pvinis !

abhitip (approved):

src/Schema/Events/ExperimentViewed.ts
```diff
@@ -0,0 +1,42 @@
...
+ * https://docs.getunleash.io/advanced/toggle_variants#variant-payload.
+ */
+export interface ExperimentViewed {
+  action: ActionType.experimentViewed
+  service: "unleash" | string
+  experiment_name: string
+  variant_name: "control" | string // usually it will be "control" or the other names of the variants
+  payload?: string
+
+  context_owner_type: OwnerType
```
mc-jones, line 38: @abhitip Did want to quickly follow up on a previous question on this attribute and confirm the requirement for this to be included with each event?

mc-jones:
@abhitip - Quickly made the changes you suggested. But did want to follow up on one previous point question on the schema.

src/Schema/Events/ExperimentViewed.ts
abhitip: Yes this should be required for every event!

src/Schema/Events/ExperimentViewed.ts
```diff
@@ -9,12 +9,12 @@ import { ActionType } from "."
...
  * A user received an experiment on Artsy.
  *
- * This schema describes events sent to Segment from [[Experiment]].
+ * This schema describes events sent to Segment from [[experimentViewed]].
  *
  *  @example
  *  ```
  *  {
- *    action: "experiment",
+ *    action: "experimentViewed",
```
pvinis, line 17: ah.. good catch 😅

artsyit:
:rocket: PR was released in [`v4.34.1`](https://github.com/artsy/cohesion/releases/tag/v4.34.1) :rocket:

PR #309, opened 2022-03-10
fix: add missing export for ExperimentViewed

mc-jones:
The type of this PR is: **Bugfix**

### Description

This PR adds the missing export to `index.ts` from #304. 


artsyit:
:rocket: PR was released in [`v4.35.0`](https://github.com/artsy/cohesion/releases/tag/v4.35.0) :rocket:

PR #322, opened 2022-04-13
[TX-227] Creating events for the payment checkout flow

daytavares:

This PR resolves [TX-227]

### Description

The payment checkout flow will change in order to add the new payment method. Therefore, we want to track clicks on the page to understand better user behaviour.

Details are described [here](https://docs.google.com/spreadsheets/d/1mEwdJ71ts-MtY7vHKzCMPAePnC_OnlpL3i4VfwTpQjI/edit?usp=sharing)

[TX-227]: https://artsyproduct.atlassian.net/browse/TX-227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -699,6 +699,62 @@ export interface ClickedPartnerCard {
 }
 
 /**
+ *  User clicks in one of the payment options on the payment pageview
+ *
+ *  This schema describes events sent to Segment from [[clickedPaymentMethod]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "clickedPaymentMethod",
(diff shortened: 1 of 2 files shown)


daytavares:
@louislecluse after you review, I would like to wait for the final check with Roberta (if design won't change).

Thanks!

louislecluse (approved):

artsyit:
:rocket: PR was released in [`v4.44.0`](https://github.com/artsy/cohesion/releases/tag/v4.44.0) :rocket:

PR #327, opened 2022-05-09
[NX-3133] Adding impulse id to buy now tap

daytavares:

This PR resolves [CO-]

### Description

- Adding `impulse id` to the tapped buy now event, so that we can track when the click happens in the conversation.
- Adding event to track when user selects edition set in the conversation (same thing we are doing for MOI)

[Ticket](https://artsyproduct.atlassian.net/browse/NX-3133)


araujobarret (approved):
Looks good, thanks 💯

artsyit:
:rocket: PR was released in [`v4.46.0`](https://github.com/artsy/cohesion/releases/tag/v4.46.0) :rocket:

PR #328, opened 2022-05-17
fix: exposes impression

dzucconi:
Exposes the impression event.

de31c3aac833f016931429329c36a6f8b2d953f6 is because some people are `--no-verify`-ing — 😠


artsyit:
:rocket: PR was released in [`v4.47.1`](https://github.com/artsy/cohesion/releases/tag/v4.47.1) :rocket:

PR #343, opened 2022-07-26
chore(deprecated-schema): Move deprecated schema from force

damassi:

### Description

We're in the process of cleaning up Force and noticed that we still have a copy of pre-cohesion events located inside, effectively duplicating the source of truth. This moves the deprecated schema out of there and into here, under the `DeprecatedSchema` folder. 

When we're ready to refactor out completely from force, we can search for the DeprecatedSchema pattern and swap out with new.


artsyit:
:rocket: PR was released in [`v4.58.0`](https://github.com/artsy/cohesion/releases/tag/v4.58.0) :rocket:

PR #345, opened 2022-08-05
[AS-3324] Adding event for balance account check

daytavares:
This PR resolves [AS-3324]

### Description

The balance check is currently performed on the client side for the buyers, so that they can switch to a different payment method right away. 

We want to measure how many times buyers run into insufficient funds from the balance check on the client side.

 **How it works now?**

Check is performed for new accounts (saved accounts are not checked). When user clicks save&continue -> checks for balance depending on how fresh of the bank account balance is. The outcomes:

- Able to get back a fresh balance that is higher than price
- Able to get back a fresh balance that is lower than price
- We don’t know (bank cannot give fresh balance or cannot have balance in the 5 seconds)

Details [here](https://docs.google.com/spreadsheets/d/1mEwdJ71ts-MtY7vHKzCMPAePnC_OnlpL3i4VfwTpQjI/edit#gid=1731346450)

[AS-3324]: https://artsyproduct.atlassian.net/browse/AS-3324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -821,6 +821,39 @@ export interface ClickedPaymentDetails {
   order_id: string
   subject: string
 }
+/**
+ * After choosing Bank Transfer, when user clicks on save & continue 
+ * on the payment page, the balance account is checked
+ *
+ *  This schema describes events sent 
+ * to Segment from [[clickedBalanceAccountCheck]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "clickedBalanceAccountCheck",
+ *    flow: "Make Offer",
+ *    context_page_owner_type: "orders-payment",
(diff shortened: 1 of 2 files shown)


daytavares:
@louislecluse I think we need a better naming for this event. What do you think?

louislecluse:
`checkedBalanceAccount`?

starsirius:
I'm OOO this week, so I'll loop in @AlicanAkyuz to help with the schema design.

daytavares:
@louislecluse I liked! I think it's easier to understand! Thanks!

louislecluse (approved):

artsyit:
:rocket: PR was released in [`v4.60.0`](https://github.com/artsy/cohesion/releases/tag/v4.60.0) :rocket:

PR #361, opened 2022-11-10
feat(CX-3105): Tracking for ContactInformation SWA step

lordkiz:

This PR resolves [CX-3105]

### Description

- With dynamic ordering of submission steps, this PR adds support to track ContactInformation step independently from ConsignmentSubmitted since ContactInformation step can now be the first step.

[CX-3105]: https://artsyproduct.atlassian.net/browse/CX-3105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/Consignments.ts
+++ b/src/Schema/Events/Consignments.ts
@@ -80,7 +80,33 @@ export interface UploadPhotosCompleted {
   action: ActionType.uploadPhotosCompleted
   context_module: ContextModule.uploadPhotos
   context_owner_type: OwnerType.consignmentFlow
-  submission_id: string
+  submission_id?: string
+  user_email: string
+  user_id?: string
+}
+
+/**
+ * First or Final step of the consignment submission flow; user provides their contact information.
+ *
+ * This schema describes events sent to Segment from [[contactInformation]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "contactInformationCompleted",
+ *    context_module: "contactInformation",
+ *    context_owner_type: "consignmentFlow",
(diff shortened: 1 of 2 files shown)


MounirDhahri (approved):
niiice

src/Schema/Events/Consignments.ts
```diff
@@ -80,7 +80,33 @@ export interface UploadPhotosCompleted {
   action: ActionType.uploadPhotosCompleted
   context_module: ContextModule.uploadPhotos
   context_owner_type: OwnerType.consignmentFlow
-  submission_id: string
+  submission_id?: string
+  user_email: string
+  user_id?: string
```
MounirDhahri, line 85: **note:** I would have set this to be `string | undefined` instead. I always prefer explicit rather than implicit in such contexts

lordkiz: Noted.

artsyit:
:rocket: PR was released in [`v4.72.0`](https://github.com/artsy/cohesion/releases/tag/v4.72.0) :rocket:

src/Schema/Events/Consignments.ts
lordkiz: @MounirDhahri Hahaha. You had merged before I could update it. No problem.

MounirDhahri: oh sorry Kizito, I just wanted to leave a note but it's definitely not blocking.

PR #363, opened 2022-11-15
feat(CX-3055):tracking hf my c onboarding part2

dariakoko:

This PR resolves [CX-3055]

### Description
modify AddCollectedArtwork event
add VisitMyCollectionevent event
add MyCollectionOnboardingScreenNUMBER events

[CX-3055]: https://artsyproduct.atlassian.net/browse/CX-3055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/MyCollection.ts
+++ b/src/Schema/Events/MyCollection.ts
@@ -26,8 +26,10 @@ export type Platform = "web" | "mobile"
  */
 export interface AddCollectedArtwork {
   action: ActionType.addCollectedArtwork
-  context_module: ContextModule.myCollectionHome
-  context_owner_type: OwnerType.myCollection
+  context_module:
+    | ContextModule.myCollectionHome
+    | ContextModule.myCollectionOnboarding
+  context_owner_type: OwnerType.myCollection | OwnerType.myCollectionOnboarding
   platform: Platform
 }
 
--- a/src/Schema/Events/HomeFeedMyCollectionOnboarding.ts
+++ b/src/Schema/Events/HomeFeedMyCollectionOnboarding.ts
@@ -75,79 +75,21 @@ export interface VisitMyCollection {
 /**
  * A user goes to the next onboarding screen.
  *
- * This schema describes events sent to Segment from [[myCollectionOnboardingScreen1]]
+ * This schema describes events sent to Segment from [[visitMyCollectionOnboardingSlide]]
  *
  *  @example
  *  ```
  *  {
- *    action: "myCollectionOnboardingScreen1",
+ *    action: "visitMyCollectionOnboardingSlide",
  *    context_screen_owner_type: "myCollectionOnboarding"
  *    context_module: "myCollectionOnboarding",
+ *    index: 1
  *  }
  * ```
  */
-export interface MyCollectionOnboardingScreen1 {
-  action: ActionType.myCollectionOnboardingScreen1
+export interface VisitMyCollectionOnboardingSlide {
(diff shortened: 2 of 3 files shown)


audrey-thenot:
LGTM! One small question, because I am not super familiar with this code: is there a way to see whether the Segment API called for each of the event is a track or a screen?

MounirDhahri (commented):
@audrey-thenot @dariakoko I am not sure I get why the action name is `myCollectionOnboardingScreenNumber` this doesn't seem like an action to me. I would have guess the action name to be something like `goToScree` or `goToSlide` or something like that. I would also expect the screen number to be simply an extra field that we can call for example `index`, this way, we can keep the same events for the future in case we add some new screens to the onboarding and we would only require a new context module.

src/Schema/Events/MyCollection.ts
```diff
@@ -26,8 +26,10 @@ export type Platform = "web" | "mobile"
  */
 export interface AddCollectedArtwork {
   action: ActionType.addCollectedArtwork
-  context_module: ContextModule.myCollectionHome
-  context_owner_type: OwnerType.myCollection
+  context_module:
+    | ContextModule.myCollectionHome
+    | ContextModule.myCollectionOnboarding
+  context_owner_type: OwnerType.myCollection | OwnerType.myCollectionOnboarding
```
MounirDhahri, line 32: **issue:** I would suggest moving the event to `/Events/HomeFeedMyCollectionOnboarding.ts` since it's part of the homefeed onboarding. Keeping it here gives the impression it's part of My Collection screen

MounirDhahri: We can of course rename it slightly

MounirDhahri (changes requested):

src/Schema/Events/MyCollection.ts
audrey-thenot: Keeping one single name for the event (only having 2 different context_owner_type) makes the pipeline update much easier, if that's possible

audrey-thenot:
> @audrey-thenot @dariakoko I am not sure I get why the action name is `myCollectionOnboardingScreenNumber` this doesn't seem like an action to me. I would have guess the action name to be something like `goToScree` or `goToSlide` or something like that. I would also expect the screen number to be simply an extra field that we can call for example `index`, this way, we can keep the same events for the future in case we add some new screens to the onboarding and we would only require a new context module.

> @MounirDhahri I actually like this idea of having only 1 screen name (like `myCollectionOnboarding`) and using a property like index to have the screen sequencing number. But I would like to track the screen view (call the screen and Segment API, not the track), like we did recently for the artwork page in My Collection (see this [ticket](https://artsyproduct.atlassian.net/browse/CX-3093)).

src/Schema/Events/MyCollection.ts
MounirDhahri: Oki sounds reasonable

MounirDhahri:
I get that, I think the issue I am raising is that the screen name is now part of the action instead of the screen name

MounirDhahri (approved):
Niiice 🙌 thanks a lot for making this change @dariakoko

artsyit:
:rocket: PR was released in [`v4.74.0`](https://github.com/artsy/cohesion/releases/tag/v4.74.0) :rocket:

PR #367, opened 2022-11-18
feat(FX-4446): track tap on curated collection on search tab

nickskalkin:

This PR resolves [FX-4446]

### Description

We are introducing a new section "Artsy Curated Collections" to a search tab. This PR adds tracking event for a tap on an Artsy curated collection in search tab.

Screenshot 2022-11-18 at 12 08 12

### DIFF
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -293,6 +293,26 @@ export interface TappedConsign {
 }
 
 /**
+ * A user taps on an Artsy curated collection on a search tab
+ *
+ * This schema describes events sent to Segment from [[tappedCuratedCollection]]
+ *
+ * @example
+ * ```
+ * {
+ *   action: "tappedCuratedCollection",
+ *   context_module: "searchTab",
+ *   slug: "trending-this-week"
+ * }
+ * ```
+ */
+export interface TappedCuratedCollection {
+  action: ActionType.tappedCuratedCollection
+  context_module: ContextModule.searchTab
+  collection_slug: string
+}
+
+/**
  * A user taps a fair card
  *
  * This schema describes events sent to Segment from [[tappedFairCard]]
--- a/src/Schema/Events/index.ts
+++ b/src/Schema/Events/index.ts
@@ -152,6 +152,7 @@ import {
   TappedConsign,
   TappedContactGallery,
   TappedCreateAlert,
+  TappedCuratedCollection,
   TappedExploreGroup,
   TappedFairCard,
   TappedFairGroup,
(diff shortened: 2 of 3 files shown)


src/Schema/Events/Tap.ts
```diff
@@ -292,6 +292,26 @@ export interface TappedConsign {
...
+ *   action: "tappedCuratedCollection",
+ *   context_module: "searchTab",
+ *   slug: "trending-this-week"
+ * }
+ * ```
+ */
+export interface TappedCuratedCollection {
+  action: ActionType.tappedCuratedCollection
+  context_module: ContextModule.searchTab
+  collection_slug: string
```
nickskalkin, line 312: Are there any other fields we want to track here? I can think of only a slug so that we know which collection performs better

src/Schema/Events/index.ts
```diff
@@ -846,6 +848,10 @@ export enum ActionType {
    * Corresponds to {@link TappedCreateAlert}
    */
   tappedCreateAlert = "tappedCreateAlert",
+  /**
+   * Corresponds to {@link TappedCuratedCollection}
+   */
+  tappedCuratedCollection = "tappedCuratedCollection",
```
nickskalkin, line 854: Any other events we should add?

src/Schema/Events/Tap.ts
```diff
@@ -292,6 +292,26 @@ export interface TappedConsign {
...
+/**
+ * A user taps on an Artsy curated collection on a search tab
+ *
+ * This schema describes events sent to Segment from [[tappedCuratedCollection]]
+ *
+ * @example
+ * ```
+ * {
+ *   action: "tappedCuratedCollection",
+ *   context_module: "searchTab",
```
dblandin, line 304: **suggestion**: I'd make the following adjustments:

- Use the existing `search` screen owner type
- Use a context module for the curated collections section
- Add `destination_*` attributes to track the tapped collection
- Add a `position` attribute

```typescript
context_screen_owner_type: ScreenOwnerType.search
context_module: ContextModule.curatedCollections
destination_screen_owner_type: ScreenOwnerType.collection
destination_screen_owner_slug: string
destination_screen_owner_id: string
position: number
```

dblandin (changes requested):
Good start! Added a comment with some suggested changes.

src/Schema/Events/Tap.ts
nickskalkin: thank you! I made changes you suggested, could you re-check?

dblandin (approved):
Nice! Looks good to me 👍🏼

artsyit:
:rocket: PR was released in [`v4.77.0`](https://github.com/artsy/cohesion/releases/tag/v4.77.0) :rocket:

PR #369, opened 2022-11-24
feat: add SelectedRecentPriceRange event for price filter

dblandin:
This commits adds a new event to track when a user engages with a new feature in the mobile app — allowing users to quickly select a previously selected price range in the artwork grid price filter screen. This list will also include a price range persisted on a user's collector profile, if available.

Example:

```typescript
{
  action: "selectedRecentPriceRange",
  context_module: "recentPriceRanges",
  context_screen_owner_type: "artworkPriceFilter",
  collector_profile_sourced: true
}
```

### DIFF
--- a/src/Schema/Events/FilterAndSort.ts
+++ b/src/Schema/Events/FilterAndSort.ts
@@ -85,3 +85,25 @@ export interface PriceDatabaseFilterParamsChanged {
   current: string
   changed: string
 }
+
+/**
+ * A user selects a recent price range from the price filter screen
+ *
+ * This schema describes events sent to Segment from [[selectedRecentPriceRange]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "selectedRecentPriceRange",
+ *    context_module: "recentPriceRanges",
+ *    context_screen_owner_type: "artworkPriceFilter",
+ *    collector_profile_sourced: true
+ *  }
+ * ```
+ */
+export interface SelectedRecentPriceRange {
+  action: ActionType.selectedRecentPriceRange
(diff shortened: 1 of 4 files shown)


src/Schema/Events/FilterAndSort.ts
```diff
@@ -85,3 +85,25 @@ export interface PriceDatabaseFilterParamsChanged {
...
+ *    collector_profile_sourced: true
+ *  }
+ * ```
+ */
+export interface SelectedRecentPriceRange {
+  action: ActionType.selectedRecentPriceRange
+  context_module: ContextModule.recentPriceRanges
+  context_screen_owner_type: OwnerType.artworkPriceFilter
+  collector_profile_sourced: boolean 
+}
```
dimatretyak, line 109: **question**: @bbaiyr should we also send the specified price range? will this information be useful?

dimatretyak (approved):
LGTM 👍

src/Schema/Events/FilterAndSort.ts
dblandin: I think the applied price range should show up in the existing `CommercialFilterParamsChanged` event so not sure if we need it here, especially because the user hasn't actually applied the filter yet.

dimatretyak: Yes, but we tracks this event only if the user clicked the "Show Results" button

artsyit:
:rocket: PR was released in [`v4.78.0`](https://github.com/artsy/cohesion/releases/tag/v4.78.0) :rocket:

PR #380, opened 2022-12-08
Change consignment_inquiry_id from bigint to number

lordkiz:
The type of this PR is: **Bugfix**

This PR resolves []

### Description
- Change consignment_inquiry_id from bigint to number because we this is typed as number on Metaphysics


artsyit:
:rocket: PR was released in [`v4.85.0`](https://github.com/artsy/cohesion/releases/tag/v4.85.0) :rocket:

PR #383, opened 2022-12-19
feat(CX-3255): Add EditedUserProfile event

MrSltun:

This PR resolves [CX-3255]

### Description
This PR adds the `EditedUserProfile` event to track updating profile events from both the app and the web

[CX-3255]: https://artsyproduct.atlassian.net/browse/CX-3255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Schema/Events/CollectorProfile.ts
```diff
@@ -0,0 +1,31 @@
...
+ *    context_screen: "collectorProfile",
+ *    context_screen_owner_type: "editProfile",
+ *    platform: "mobile"
+ *  }
+ * ```
+ */
+export interface EditedUserProfile {
+  action: ActionType.editedUserProfile
+  context_screen: ContextModule.collectorProfile
+  context_screen_owner_type: OwnerType.editProfile
```
MrSltun, line 29: **Suggestion:** Since we are using the same event for both mobile and web, and the keyword `screen` indicates that it's from mobile, should we change these to `context_module...` instead?

MounirDhahri (approved):

audrey-thenot (approved):

src/Schema/Events/CollectorProfile.ts
audrey-thenot: Both sound OK to me. We can use the context_module is you think that this makes more sense.

MounirDhahri: @audrey-thenot what do you think?

MrSltun: I believe it would be okay to have it as is now, I just wanted to check with Audrey about it :) will merge now!

artsyit:
:rocket: PR was released in [`v4.87.0`](https://github.com/artsy/cohesion/releases/tag/v4.87.0) :rocket:

src/Schema/Events/CollectorProfile.ts
MounirDhahri: Sure - please feel free to reach out to Audrey personally if blocked

PR #391, opened 2023-01-17
chore: add context module to ExperimentViewed

MounirDhahri:

### Description

Add context module to ExperimentViewed.


MounirDhahri:
#squashongreen

artsyit:
:rocket: PR was released in [`v4.94.0`](https://github.com/artsy/cohesion/releases/tag/v4.94.0) :rocket:

PR #398, opened 2023-02-03
[AS-3677] CMS conversation page events

daytavares:

This PR resolves [CO-]

### Description
With the conversations revamp, we want to migrate server events to client events.

Our approach to cover the previous server events is:
* we will track most of clicks on the conversation page with click event (`volt_client_production.click`), probably another NX ticket. 
* we will create some specific events that are relevant to be separated

List of server events to be migrated:

- **conversations_index** -> [tracked using pageviews](https://artsyproduct.atlassian.net/browse/AS-3676)
- **conversations_sent** -> tracked with the new event in this PR `ClickedConversationsFilter`
- **conversations_new_inquiries** -> tracked with the new event in this PR `ClickedConversationsFilter`
- **conversations_show** -> will be tracked with click
- **conversations_update_seller_outcome** -> tracked with the new event in this PR `ClickedDismissInquiry`
- **conversations_messages_create** -> will be tracked with click
- **conversations_messages_mark_as_spam** -> tracked with the new event in this PR `ClickedMarkSpam`
- **artworks_update** -> tracked with the new event in this PR `ClickedDismissInquiry`

More details [here](https://docs.google.com/spreadsheets/d/1pc1tlSlNuVmRfNJBgvwb3CBxo5gH6ewuL0-Qle4bwOA/edit#gid=1140630574)

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -1423,8 +1423,8 @@ export interface ClickedOnPriceDisplayDropdown {
 }
 
 /**
- * Partner clicks on dismiss inquiry on the conversations page in CMS. 
- * A pop up will open and then they can click on: Select a reason, Cancel, Dismiss inquiry 
+ * Partner clicks on dismiss inquiry modal on the conversations page in CMS.
+ * They can click on: Select a reason, Cancel, Dismiss inquiry 
  *
  * This schema describes events sent to Segment from [[clickedDismissInquiry]]
  *
@@ -1453,8 +1453,8 @@ export interface ClickedOnPriceDisplayDropdown {
   partner_id: string
 }
 /**
- * Partner clicks on mark as spam on the conversations page in CMS. 
- * A pop up will open and then they can click on: Cancel or Delete and Mark as Spam
+ * Partner clicks on mark as spam modal on the conversations page in CMS. 
+ * They can click on: Cancel or Delete and Mark as Spam
  *
  * This schema describes events sent to Segment from [[clickedMarkSpam]]
  *
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -1395,3 +1395,88 @@ export interface ClickedOnPriceDisplayDropdown {
   context_page_owner_id?: string
   label: string
 }
+
+/**
+ * A Partner selects a filter on the conversations page in CMS.
+ *
+ * This schema describes events sent to Segment from [[selectedConversationsFilter]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    label: [selected filter all, selected filter new, selected filter replied],
+ *    context_module: "conversations",
+ *    context_page_owner_type: "conversation",
+ *    context_page_owner_id: "60de173a47476c000fd5c4cc"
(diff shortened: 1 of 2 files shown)


louislecluse:
Thanks! I'm onboard with the approach.
For the question around "dismiss inquiry and mark as spam", can you provide screenshots (you can send on slack)? I'm not familiar enough with the experience to help.

daytavares:
[Notion page](https://www.notion.so/artsy/Tracking-Monitoring-f27c1f5a758141838a0f2942e6fd191e)

daytavares:
Screenshot 2023-02-03 at 15 26 03
Screenshot 2023-02-03 at 15 26 29
Screenshot 2023-02-03 at 15 26 40

daytavares:
cc @artsy/negotiate-devs

louislecluse:
Thanks for providing these details! I recommend tracking a click for the buttons on the conversation page and then tracking dismiss inquiry and mark as spam when users interact with the modal.
Otherwise it might be difficult to differentiate when user clicks the buttons on the conversations page versus inside the modal.
For example, if you look at the inquiry flow, we track different events when collector clicks "contact gallery" versus going through the flow and then sending the inquiry.

lidimayra:
Thank you, @daytavares !! 🎉 
I have just one question. It's regarding this one:
> artworks_update -> tracked with the new event in this PR ClickedDismissInquiry

Not clear to me how will we track if the artwork was updated or not using this one. Should it be a separate event or am I missing something?

abhitip:
Thanks dayane! The only thing I'm not seeing here is our migration plan in fulcrum. If we're changing event names, we will need to account for that in our pipeline. Happy to help write this plan!

daytavares:
@lidimayra the event `artworks_update` is on the artwork form? Because I tested and when I chose "This artwork is no longer available", I couldn't see this option to mark as sold. 
I thought that using the event `ClickedDismissInquiry` and the reason "This artwork is no longer available", we would know the artwork is sold.

daytavares:
@abhitip I have a [ticket](https://artsyproduct.atlassian.net/browse/AS-3754) for it 😊

daytavares:
@louislecluse also prefer the approach of only tracking the modal clicks 😊

lidimayra:
> @lidimayra the event `artworks_update` is on the artwork form? Because I tested and when I chose "This artwork is no longer available", I couldn't see this option to mark as sold. I thought that using the event `ClickedDismissInquiry` and the reason "This artwork is no longer available", we would know the artwork is sold.

Ohh that's true. We haven't yet implemented that functionality on V2.

I had noted that one down because I was using V1 as reference. On V1, in case the partner chooses "This artwork is no longer available", they have the possibility of updating the availability as well:
set_as_sold

I'm not entirely sure we'll have the same feature in V2 though, especially given some [recent discussions](https://artsy.slack.com/archives/C9YNS4X32/p1675334947122859?thread_ts=1675334184.995699&cid=C9YNS4X32).

So, I guess it might make sense to also not include tracking for this feature at this point (and think about this when we have more clarity on the expected behavior)

artsyit:
:rocket: PR was released in [`v4.100.0`](https://github.com/artsy/cohesion/releases/tag/v4.100.0) :rocket:

PR #399, opened 2023-02-20
Fixing CMS conversation - ClickedConversationsFilter

daytavares:

This PR resolves [CO-]

### Description

We created previously events to track events on CMS conversations v2.

For `ClickedConversationsFilter` though we won't have `artwork_id` as we are in the list of conversations and not in a specific conversation.


lidimayra (approved):
Thank you!! 🙌 💜

artsyit:
:rocket: PR was released in [`v4.101.0`](https://github.com/artsy/cohesion/releases/tag/v4.101.0) :rocket:

PR #416, opened 2023-03-10
feat: add item viewed

MounirDhahri:

This PR resolves [CX-3465]

### Description

Add item viewed event

[CX-3465]: https://artsyproduct.atlassian.net/browse/CX-3465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Schema/Events/Search.ts
```diff
@@ -208,4 +208,4 @@ export interface SelectedSearchSuggestionQuickNavigationItem {
   context_module?: ContextModule
   destination_path: string
   label: "Artworks" | "Auction Results"
-}
\ No newline at end of file
+}
```
MounirDhahri, line 211: prettier did this 😄

PR #419, opened 2023-03-20
[AS-3848] Adding new event for mark as sold

daytavares:

This PR resolves [CO-]

### Description
Add tracking for the mark as sold checkbox in the Dismiss Inquiry modal on CMS conversation page.

When the checkbox is selected (for the work_unavailable reason) we mark the artwork as sold after dismissing the inquiry.


louislecluse (approved):
lgtm

artsyit:
:rocket: PR was released in [`v4.120.0`](https://github.com/artsy/cohesion/releases/tag/v4.120.0) :rocket:

PR #422, opened 2023-03-29
chore: add tracking event for clicking publish button in artwork form 

JanaeHijaz:
The type of this PR is: Chore

### Description

This PR adds a new event to cohesion for the publish button, resolving the first part of [PX-5499] 
(additional PR for volt changes *add link here*)

[PX-5499]: https://artsyproduct.atlassian.net/browse/PX-5499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/index.ts
+++ b/src/Schema/Events/index.ts
@@ -53,8 +53,8 @@ import {
   ClickedGalleryGroup,
   ClickedLoadMore,
   ClickedMainArtworkGrid,
-  ClickedMarkSpam,
   ClickedMarkSold,
+  ClickedMarkSpam,
   ClickedNavigationTab,
   ClickedOfferActions,
   ClickedOfferOption,
@@ -73,6 +73,7 @@ import {
   ClickedPaymentDetails,
   ClickedPaymentMethod,
   ClickedPromoSpace,
+  ClickedPublish,
   ClickedSelectShippingOption,
   ClickedShippingAddress,
   ClickedShowGroup,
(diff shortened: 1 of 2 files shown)


ansor4 (approved):

src/Schema/Events/index.ts
```diff
@@ -579,6 +581,10 @@ export enum ActionType {
    * Corresponds to {@link ClickedOnPriceDisplayDropdown}
    */
   clickedOnPriceDisplayDropdown = "clickedOnPriceDisplayDropdown",
+  /**
+   * Corresponds to {@link ClickedPublish}
+   */
+  clickedPublish = "clickedPublish",
```
ansor4, line 587: I'm actually not sure if this is correct. Looking at the [spreadsheet](https://docs.google.com/spreadsheets/d/124_JBqIvgUGB7t3XcJeEew2Bvd624Y6qyQgYxK0V-3E/edit#gid=454969903) referenced in the ticket, I think this may need to be a different value?

ansor4 (changes requested):

ansor4:
After talking with @louislecluse, let's do the following:
- We can go with the metadata:
  - action: "Click"
  - label: "Publish"
  - context module: "Artworks - artworks detail page"
  - artwork_id
- Since we have this action already, we don't need to add anything to cohesion. Let's close this PR and open a Volt one
  - We can look at what the publish from artwork list button does

Thanks!

nspinazz89 (approved):
This looks good to me! I would want to get final confirmation from @louislecluse that this is indeed how we want to do this!

louislecluse:
Thanks! We may not always track the artwork id with that implementation. (for example if we're on a page where the context is not artwork). 
So we should add an artwork id property.

Otherwise looks good!

artsyit:
:rocket: PR was released in [`v4.125.0`](https://github.com/artsy/cohesion/releases/tag/v4.125.0) :rocket:

PR #430, opened 2023-04-28
chore: make sale id optional for IDV

mdole:
The type of this PR is: **CHORE**

Jira ticket: https://artsyproduct.atlassian.net/browse/PRO-107

### Description

Since we want to start sending these events when uses click from the profile page, which does not have an associated auction, `sale_id` should become optional.

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -1123,7 +1123,7 @@ export interface ClickedVerifyIdentity {
   context_page_owner_type: PageOwnerType
   context_page_owner_id?: string
   context_page_owner_slug?: string
-  sale_id: string
+  sale_id?: string
   subject: string
 }
 


src/Schema/Events/Click.ts
```diff
@@ -1123,7 +1123,7 @@ export interface ClickedVerifyIdentity {
   context_page_owner_type: PageOwnerType
   context_page_owner_id?: string
   context_page_owner_slug?: string
-  sale_id: string
+  sale_id?: string
```
mdole, line 1126: @daytavares should `subject` also be optional? Not super clear on what it's used for 🤔

daytavares (approved):

src/Schema/Events/Click.ts
mdole: OK thanks! Then let's keep it as required

artsyit:
:rocket: PR was released in [`v4.125.1`](https://github.com/artsy/cohesion/releases/tag/v4.125.1) :rocket:

PR #431, opened 2023-05-12
Event to track errors

louislecluse:
@iskounen Let me know if this works!

### DIFF
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -94,7 +94,8 @@ export interface TooltipViewed {
  *    context_owner_id: "order-id",
  *    title: "This is an error",
  *    message: "Error message",
- *    error_code: "charge_authorization_failed"
+ *    error_code: "charge_authorization_failed",
+ *    flow: "user submitting payment"
  *  }
  * ```
  *
@@ -105,5 +106,6 @@ export interface ErrorMessageViewed {
   context_owner_type: OwnerType
   title: string
   message: string
-  error_code: string
+  error_code?: string
+  flow: string
 }
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -80,3 +80,28 @@ export interface TooltipViewed {
   context_owner_type: PageOwnerType
   type: string
 }
+
+/**
+ * A user sees a an error message
+ *
+ * This schema describes events sent to Segment from [[ErrorMessageViewed]].
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "errorMessageViewed",
+ *    context_owner_type: "orders",
+ *    context_owner_id: "order-id",
+ *    label: "This is an error",
+ *    error_code: "charge_authorization_failed"
+ *  }
+ * ```
+ *
+ */
+export interface ErrorMessageViewed {
+  action: ActionType.errorMessageViewed
+  context_owner_id: string
+  context_owner_type: OwnerType
+  label: string
+  error_code: string
+}
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -92,7 +92,8 @@ export interface TooltipViewed {
  *    action: "errorMessageViewed",
  *    context_owner_type: "orders",
  *    context_owner_id: "order-id",
- *    label: "This is an error",
+ *    title: "This is an error",
+ *    message: "Error message",
  *    error_code: "charge_authorization_failed"
  *  }
  * ```
@@ -102,6 +103,7 @@ export interface ErrorMessageViewed {
   action: ActionType.errorMessageViewed
   context_owner_id: string
   context_owner_type: OwnerType
-  label: string
+  title: string
+  message: string
   error_code: string
 }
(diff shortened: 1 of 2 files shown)


louislecluse:
For the `label` property, I'd like to track the message we output to the user.

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -80,3 +80,28 @@ export interface TooltipViewed {
...
+ *  }
+ * ```
+ *
+ */
+export interface ErrorMessageViewed {
+  action: ActionType.errorMessageViewed
+  context_owner_id: string
+  context_owner_type: OwnerType
+  label: string
+  error_code: string
```
iskounen, line 106: We won't always have an error code when these errors occur. Could we make this optional and provide something like the name of the route where the error was raised?

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -80,3 +80,28 @@ export interface TooltipViewed {
...
+ *    error_code: "charge_authorization_failed"
+ *  }
+ * ```
+ *
+ */
+export interface ErrorMessageViewed {
+  action: ActionType.errorMessageViewed
+  context_owner_id: string
+  context_owner_type: OwnerType
+  label: string
```
iskounen, line 105: We have a title and a message field, both of which are interested. Could we have two fields for those, or should we smoosh them together as one label?

src/Schema/Events/ImpressionTracking.ts
iskounen: Is it better to use a 'generic_error_code' than a null if none is available?

src/Schema/Events/ImpressionTracking.ts
iskounen: Instead of using the error message as the label, we decided to describe the user action that led to the error (ex. user submitted payment method) so that we can differentiate the places where an error dialog can be displayed in one route.

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -80,3 +80,28 @@ export interface TooltipViewed {
...
+  error_code: string
+}
```
iskounen, line 107: The error dialog is displayed on several routes in the orders app in Force, so we would like to differentiate which route the error message is displayed on (ie. shipping, payment, counteroffer, etc.).

Could we have a separate field for the route?

src/Schema/Events/ImpressionTracking.ts
louislecluse: How do you call this "action" field?

louislecluse: We can have a property for each: message / title, I'll add

src/Schema/Events/ImpressionTracking.ts
louislecluse: Should be on the context page path that we record always through the event!

src/Schema/Events/ImpressionTracking.ts
iskounen: It would be a placeholder that we send instead of `null` when an error code is not available.

src/Schema/Events/ImpressionTracking.ts
iskounen: That works too. We thought that a field to describe the action that caused an error would be better than the title/message pair because:

1. The same title/message pair can be shown to users for different actions on the same page, so mapping an event to the action will allow us to pinpoint the action that caused it.
2. TX is planning to revise the copy of existing error dialogs in the checkout flow, so we could lose track of the events that we have collected if we rely on the title/message pair.

I'm not sure what I would call this field. I did a quick search through the existing Cohesion types and didn't find one that matched. Maybe something like `context_user_action`?

src/Schema/Events/ImpressionTracking.ts
iskounen: That works 👍

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -80,3 +80,30 @@ export interface TooltipViewed {
...
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "errorMessageViewed",
+ *    context_owner_type: "orders",
+ *    context_owner_id: "order-id",
+ *    title: "This is an error",
+ *    message: "Error message",
+ *    error_code: "charge_authorization_failed"
```
rquartararo, line 97: Can we add a "route" field to track which route the error occurs on, such as "payment", "shipping", "offer", "review", etc?

rquartararo: Oh oops, I didn't see the conversation! that sounds good thanks!

src/Schema/Events/ImpressionTracking.ts
louislecluse: Got it! We can leave null I think!

src/Schema/Events/ImpressionTracking.ts
louislecluse: I can add an `action` property to record also that specific action.

iskounen (approved):
@louislecluse thanks a lot for your help with this, and for helping us talk through the properties that we need!

artsyit:
:rocket: PR was released in [`v4.127.0`](https://github.com/artsy/cohesion/releases/tag/v4.127.0) :rocket:

PR #433, opened 2023-05-25
feat: add events schema for search pills

rajsam003:
The type of this PR is: Feature

Addresses [FX-4780]

### Description

Adds events schema for search pill and its chevron icons

[FX-4780]: https://artsyproduct.atlassian.net/browse/FX-4780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Schema/Events/index.ts
```diff
@@ -915,6 +921,18 @@ export enum ActionType {
...
+   */
+  selectedPillFromSearch = "selectedPillFromSearch",
+  /**
+   * Corresponds to {@link ClickedOnRightChevronIconOnSearchPills}
+   */
+  clickedOnRightChevronIconOnSearchPills = "clickedOnRightChevronIconOnSearchPills",
+  /**
+   * Corresponds to {@link ClickedOnLeftChevronIconOnSearchPills}
+   */
+  clickedOnLeftChevronIconOnSearchPills = "clickedOnLeftChevronIconOnSearchPills",
```
dimatretyak, line 935: **question**: do we really want (and need) to track this action?

src/Schema/Events/Search.ts
```diff
@@ -88,6 +88,86 @@ export interface SearchedWithNoResults {
...
+ */
+export interface SelectedPillFromSearch {
+  action: ActionType.selectedPillFromSearch
+  context_module: ContextModule
+  context_owner_type: PageOwnerType
+  context_owner_id?: string
+  context_owner_slug?: string
+  destination_owner_type?: PageOwnerType
+  query: string
+  pill: string
```
gkartalis, line 116: **question(non-blocking):** Should we maybe define a type for pill names? 🤔

src/Schema/Events/Search.ts
```diff
@@ -88,6 +88,86 @@ export interface SearchedWithNoResults {
...
+ *    action: "selectedPillFromSearch",
+ *    context_module: "header",
+ *    context_owner_type: "home",
+ *    destination_owner_type: "search",
+ *    query: "andhuwe"
+ *    pill: "artwork"
+ *  }
+ * ```
+ */
+export interface SelectedPillFromSearch {
```
dimatretyak, line 108: **question**: @abhitip should we use the `tappedNavigationTab` event when clicking on pills, as it happens in [Eigen](https://github.com/artsy/eigen/blob/192d1aa85d7435406f5bb51213a8fe28fa3382ab/src/app/Scenes/Search/Search2.tsx#L206-L213)?

dimatretyak (approved):

src/Schema/Events/index.ts
rajsam003: I can remove it. but what do you think @abhitip ?

src/Schema/Events/Search.ts
gkartalis: I was thinking since we have specific pill names it could be something like:
```ts
pill: Pill

type Pill = "Artist" | "Artwork" 
``` 

etc, but I checked eigen and it doesn't follow the same so 🤷‍♂️ feel free to ignore

src/Schema/Events/Search.ts
rajsam003: What is `tappedNavigationTab`?

rajsam003: Is it similar to - what pill was active when a user taps a result ?

src/Schema/Events/Search.ts
dblandin: I agree that we should type the allowed values here in cohesion. I think we can probably use the `key` values [here](https://github.com/artsy/eigen/blob/main/src/app/Scenes/Search/constants.ts#L25-L83) as an initial set.

So something like:

```typescript
type PillTypes = "top" | "artwork" | "artist" | "article" | "sale" | "artist_series" | "marketing_collection" | "fair" | "partner_show" | "PartnerGallery"
```

A little strange with the different casing between `partner_show` and `PartnerGallery` but might be fine as long as it's well-documented in this way.

src/Schema/Events/Search.ts
dblandin: Oh, interesting, I guess we use `subject` to capture that displayed text of the navigation tab ([source](https://github.com/artsy/cohesion/blob/1dcd0355091b3af34d478b26c1a1c1e5bed089ae/src/Schema/Events/Tap.ts#L559-L567)). Used here [in Force too](https://github.com/artsy/force/blob/main/src/Apps/Search/Components/NavigationTabs.tsx#L78-L86).

So yeah, maybe we can reuse that event and drop the idea of typing the allowed "pills" which are more susceptible to change if it's the display name we pass through.

abhitip: Oh yes let's use the same event if possible! That naming is a little confusing with the "navigation" but it's okay

src/Schema/Events/index.ts
abhitip: Agreed, I think we can remove the chevron events!

src/DeprecatedHelpers/Search/SearchedWithNoResults.ts
```diff
@@ -6,6 +6,7 @@ export const searchedWithNoResults = (
 ): SearchedWithNoResults => {
   return {
     action: ActionType.searchedWithNoResults,
+    activePill: props.activePill,
```
dblandin, line 9: I know in [a previous review comment](https://github.com/artsy/cohesion/pull/433#discussion_r1205528162) we explored having a `PillType` and being more explicit with a property like this but to remain consistent with other surfaces, as as [I brought up in Slack](https://artsy.slack.com/archives/C9SATFLUU/p1685026012585709?thread_ts=1685026008.427369&cid=C9SATFLUU), I think we should use the existing `context_module` pattern and add any additional `[type]Tab` ContextModule properties if we need to.

So far with have the following src/Schema/Values/ContextModule.ts to use:

```
artistSeriesTab
artworksTab
auctionTab
fairTab
galleryTab
showTab
topTab
```

It looks like the only changes we need are to add `marketingCollectionTab` and `articleTab` ContextModule options and continue to use the existing `tappedNavigationTab` and `selectedItemFromSearch` events.

dimatretyak: +1 to Devon's comment

src/Schema/Events/Search.ts
```diff
@@ -80,6 +84,36 @@ export interface SelectedItemFromSearch {
...
+ *    activePill: "artwork"
+ *    context_module: "header",
+ *    context_owner_type: "home",
+ *    destination_owner_type: "search",
+ *    query: "andhuwe"
+ *    pill: "artwork"
+ *  }
+ * ```
+ */
+export interface SelectedPillFromSearch {
```
dimatretyak, line 114: this event is duplicated by [an existing one](https://github.com/artsy/cohesion/blob/acf10eddd3f18226704c5688da5fd3bbe4c8ced4/src/Schema/Events/Tap.ts#L559-L567)

src/DeprecatedHelpers/Search/SearchedWithNoResults.ts
rajsam003: Then we dont need the `activePill` field as we already have `context_module`

dblandin: Yes, exactly. I think the only changes expected of this PR at this time are new allowed values in the `ContextModule` [file here](https://github.com/artsy/cohesion/blob/41c58e113df31ca8459e014f1f1b468642fdaf0a/src/Schema/Values/ContextModule.ts#L7).

src/Schema/Events/Search.ts
```diff
@@ -38,7 +38,7 @@ export interface FocusedOnSearchInput {
  *  ```
  *  {
  *    action: "selectedItemFromSearch",
- *    context_module: "priceEstimate",
+ *    context_module: "topTab",
```
dblandin, line 41: **suggestion**: `priceEstimate"` is still a valid ContextModule value and probably makes more sense paired with the rest of the example attributes here so I might remove this change.

src/Schema/Events/Search.ts
```diff
@@ -71,7 +71,7 @@ export interface SelectedItemFromSearch {
  *  ```
  *  {
  *    action: "searchedWithNoResults",
- *    context_module: "header",
+ *    context_module: "topTab",
```
dblandin, line 74: Good reminder that the more generic `header` ContextModule will be less often used in favor of these more specific pill/tab values.

cc/ @abhitip

dblandin (approved):
One small suggestion but overall LGTM!

artsyit:
:rocket: PR was released in [`v4.128.0`](https://github.com/artsy/cohesion/releases/tag/v4.128.0) :rocket:

PR #437, opened 2023-06-12
chore: PX-5596 adding new event for duplicate artwork functionality

Mitch-Henson:

This PR is required to complete [PX-5596]

### Description

This PR adds a new event to Cohesion to track when an artwork is being duplicated in CMS.

[PX-5596]: https://artsyproduct.atlassian.net/browse/PX-5596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -1594,16 +1594,16 @@ export interface ClickedOnMakeOfferCheckbox {
  *  {
  *    action: "clickedOnDuplicateArtwork",
  *    context_module: "voltArtworksEdit",
- *    context_page_owner_type: "artwork",
- *    context_page_owner_id: "60de173a47476c000fd5c4cc"
- *    label: ["duplicate", "cancel"]
+ *    original_artwork_id: "60de173a47476c000fd5c4cc"
+ *    duplicate_artwork_id: "6487822214765c000d34cbe8"
+ *    label: ["duplicate", "cancel", "open modal"]
  *  }
  * ```
  */
 export interface ClickedOnDuplicateArtwork {
   action: ActionType.clickedOnDuplicateArtwork
   context_module: ContextModule
-  context_page_owner_type: PageOwnerType
-  context_page_owner_id?: string
+  original_artwork_id: string
+  duplicate_artwork_id?: string
   label: string
 }
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -1595,8 +1595,8 @@ export interface ClickedOnMakeOfferCheckbox {
  *    action: "clickedOnDuplicateArtwork",
  *    context_module: "voltArtworksEdit",
  *    original_artwork_id: "60de173a47476c000fd5c4cc"
- *    duplicate_artwork_id: "6487822214765c000d34cbe8"
- *    label: ["duplicate", "cancel", "open modal"]
+ *    label: "Duplicate"
+ *    flow: ["artworksList", "duplicateModal"]
  *  }
  * ```
  */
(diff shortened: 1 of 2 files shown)


jpotts244 (approved):
LGTM!

louislecluse:
Thanks!
I would like to track the id of the original duplicated work (and ideally the new work - but that's optional).

We need to track more on the properties to do that. From what I understand the duplicate button will be on the artworks list page. So `context_page_owner_id` will not give us the id of the artwork.
So we can add a new property like `original_artwork_id` or something like that?

src/Schema/Events/Click.ts
```diff
@@ -1583,3 +1583,27 @@ export interface ClickedOnMakeOfferCheckbox {
...
+ *    duplicate_artwork_id: "6487822214765c000d34cbe8"
+ *    label: ["duplicate", "cancel", "open modal"]
+ *  }
+ * ```
+ */
+export interface ClickedOnDuplicateArtwork {
+  action: ActionType.clickedOnDuplicateArtwork
+  context_module: ContextModule
+  original_artwork_id: string
+  duplicate_artwork_id?: string
```
jpotts244, line 1607: Ah yes, interesting point about the `duplicate_artwork_id`. If we do truly want it, it the event will have to be fired after the response comes back from Gravity which kinda makes this event not really a "click" type event, right?

louislecluse:
On the modal:
I don't think we need to track "cancel", but makes sense to track both "duplicate" clicks. The `label` property is used for what's written in the UI (eg: "Duplicate Artwork"), so I'd rather use another property like `flow` to record either `artworksList` of `duplicateModal`

No need to track the new work id.

JanaeHijaz:
> I can definitely add the new property `original_artwork_id`!
> 
> In terms of getting the ID of the new work it isn't going to be as simple. We only know the ID of the new work once the API has returned which means that we have to call the `trackEvent` _after_ the API returns, which isn't ideal. I can add the property in as nullable though and see how well the implementation works? There is a question of what to do if the backend errors though. Do we still make the `trackEvent` call even though it wasn't successful?
> 
> Lastly, I have a question as I saw that we want to track when the duplicate button is first clicked which opens the modal. Do we want to differentiate between clicking that button and clicking the duplicate button inside of the modal (which actually triggers the duplication call to the backend). If so, I could update the label to have three different values: "duplicate", "cancel" and "open modal"
> 
> Screenshot 2023-06-13 at 11 24 18 Screenshot 2023-06-13 at 11 23 32

@Mitch-Henson @louislecluse there is already the field `duplicated_from_artwork_id` that we've added to the artwork model. Is that what you're referring to? I don't see why we would need to create a new field such as `original_artwork_id`

JanaeHijaz:
> Thanks! I would like to track the id of the original duplicated work (and ideally the new work - but that's optional).
> 
> We need to track more on the properties to do that. From what I understand the duplicate button will be on the artworks list page. So `context_page_owner_id` will not give us the id of the artwork. So we can add a new property like `original_artwork_id` or something like that?

@Mitch-Henson @louislecluse there is already the field `duplicated_from_artwork_id` that we've added to the artwork model. Is that what you're referring to? I don't see why we would need to create a new field such as `original_artwork_id`.

jpotts244 (approved):

artsyit:
:rocket: PR was released in [`v4.131.0`](https://github.com/artsy/cohesion/releases/tag/v4.131.0) :rocket:

PR #440, opened 2023-07-03
Add SearchedWithResults event

nickskalkin:

### Description

New search bar implementation [uses](https://github.com/artsy/force/blob/114efdb251f8d4719281455e2efbe8a937b78305/src/Components/Search/NewSearch/NewSearchBarInput.tsx#L90) deprecated search events, like:
- DeprecatedSchema.ActionType.SearchedAutosuggestWithResults
- DeprecatedSchema.ActionType.SearchedAutosuggestWithoutResults

I've added a new event for searching **with** results ([there is](https://github.com/artsy/cohesion/blob/60456c53822b8c29e3068fe53e70a43ff839d051/src/Schema/Events/Search.ts#L81) an event for searching **without** results in the schema already) In order to not drag along deprecated events.

### DIFF
--- a/src/Schema/Events/Search.ts
+++ b/src/Schema/Events/Search.ts
@@ -89,6 +89,32 @@ export interface SearchedWithNoResults {
 }
 
 /**
+ * A user searches with results
+ *
+ * This schema describes events sent to Segment from [[searchedWithResults]]
+ *
+ *  @example
+ *  ```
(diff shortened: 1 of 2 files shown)


dblandin (commented):
LGTM 👍🏼

dblandin (approved):

PR #443, opened 2023-07-17
[AS-4029] Adding new event to track address validation

daytavares:
The type of this PR is: **Feature**

This PR resolves [EMI-1285](https://artsyproduct.atlassian.net/browse/EMI-1285)

### Description

This PR adds a new event to Cohesion to track when an user clicks on the buttons on the address verification modals

More details [here](https://www.notion.so/artsy/Address-auto-completion-and-verification-d3a89ec48845408f9d3a2724ce4e09b8?pvs=4#a6d4cf8409504609bdfdb309cd930de3)

### PR Checklist (tick all before merging)

- [ ] If I've added a new file to the tree I've exported it from the common [`index.ts`](https://github.com/artsy/cohesion/blob/main/src/Schema/Events/index.ts)
- [ ] I've added comments with examples for any new interfaces and ensured that they're in the docs
- [ ] No platform-specific terminology has been used outside of `click` and `tap` (platform is inferred by the DB storing events)

[EMI-1285]: https://artsyproduct.atlassian.net/browse/EMI-1285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #445, opened 2023-07-19
(feat): add modalType option to address validation event

TMcMeans:

The goal of this PR is to add an option and subject property to the `ValidationAddressViewed` tracking event so that analytics can differentiate which modal (address suggestions or general verification error) is viewed by user. We will better be able to analyze triggers here, per @starsirius's notes on this PR: https://github.com/artsy/cohesion/pull/443

@daytavares please let me know if I have things in the right place or if changes are requested. My goal was to make some quick edits to the schema so that we can keep progressing on analytics tracking.

### DIFF
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -123,7 +123,9 @@ export interface ErrorMessageViewed {
  *    context_page_owner_type: "orders-shipping",
  *    context_page_owner_id: "57e60c68-a198-431e-8a02-6ecb01e3a99b",
  *    user_id: "61bcda16515b038ce5000104"
- *    flow: "user adding shipping address"
+ *    flow: "user adding shipping address",
+ *    subject: "Check your delivery address",
+      option: "review and confirm",
  *  }
  * ```
  *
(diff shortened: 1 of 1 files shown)


src/Schema/Events/ImpressionTracking.ts
```diff
@@ -123,7 +123,9 @@ export interface ErrorMessageViewed {
  *    context_page_owner_type: "orders-shipping",
  *    context_page_owner_id: "57e60c68-a198-431e-8a02-6ecb01e3a99b",
  *    user_id: "61bcda16515b038ce5000104"
- *    flow: "user adding shipping address"
+ *    flow: "user adding shipping address",
+ *    subject: "Check your delivery address",
+      option: "review and confirm",
```
starsirius, line 128: Thank you! Here is the [relevant thread](https://artsy.slack.com/archives/C02JHHHKP5K/p1689837587009359?thread_ts=1689786251.379909&cid=C02JHHHKP5K) to finalize the new properties.

daytavares:
It looks good! Thanks!

artsyit:
:rocket: PR was released in [`v4.135.0`](https://github.com/artsy/cohesion/releases/tag/v4.135.0) :rocket:

PR #448, opened 2023-08-18
CMS alerts events tracking

louislecluse:
Tracking for CMS alerts!

I added some more generic events for Nav bar clicks and upload artworks that can be re-used elsewhere.

PR #450, opened 2023-08-31
fix: saved alerts events

lidimayra:

[Slack thread](https://artsy.slack.com/archives/C02BC3HEJ/p1693410047157529)
Relates to https://github.com/artsy/cohesion/pull/448

### Description

Fixes issues that are leading the build to fail


tam-kis (approved):
Makes sense, the error message was also pointing to this - thanks! 👍

artsyit:
:rocket: PR was released in [`v4.137.0`](https://github.com/artsy/cohesion/releases/tag/v4.137.0) :rocket:

PR #457, opened 2023-09-13
[AS-4137] Address auto completion tracking

daytavares:

This PR resolves [AS-4137]

### Description

This PR adds a new event to Cohesion to track user interactions with shipping address auto completion

More details [here](https://docs.google.com/spreadsheets/d/1c65vugkF2C4YuXwBXno8aqsBh4B2fTwxiLmgY_MhjWc/edit#gid=0)

[AS-4137]: https://artsyproduct.atlassian.net/browse/AS-4137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/Search.ts
+++ b/src/Schema/Events/Search.ts
@@ -237,27 +237,29 @@ export interface SelectedSearchSuggestionQuickNavigationItem {
 }
 
 /**
- * A user focuses out an address line box when there is input
- *
- * This schema describes events sent to Segment from [[focusedOutAddresssBox]]
+ * User fills shipping address and has/or not auto completion with results
+ * 
+ * This schema describes events sent to Segment from [[addressAutoCompletionResult]]
  *
  *  @example
  *  ```
  *  {
- *    action: "focusedOutAddresssBox",
+ *    action: "addressAutoCompletionResult",
  *    context_module:"OrdersShipping",
  *    context_page_owner_type: "orders-shipping",
  *    context_page_owner_id: "57e60c68-a198-431e-8a02-6ecb01e3a99b",
  *    input: "Weserstr."
+ *    suggested_addresses_results: 3 
  *  }
  * ```
  */
- export interface FocusedOutAddresssBox {
-  action: ActionType.focusedOutAddresssBox
+ export interface AddressAutoCompletionResult {
+  action: ActionType.addressAutoCompletionResult
   context_module: ContextModule
   context_owner_type: PageOwnerType
   context_owner_id?: string
   input: string
+  suggested_addresses_results: number 
 }
 
 /**
@@ -274,8 +276,6 @@ export interface SelectedSearchSuggestionQuickNavigationItem {
  *    context_page_owner_id: "57e60c68-a198-431e-8a02-6ecb01e3a99b",
  *    input: "Wes"
  *    item: "Weserstr."
- *    fetches: 3
- *    results: 2
  *  }
  * ```
  */
@@ -286,6 +286,29 @@ export interface SelectedSearchSuggestionQuickNavigationItem {
   context_owner_id?: string
   input: string
   item: string
-  fecthes: number
-  results: number
+}
+
+/**
+ * User edits a suggested address
+ *
+ * This schema describes events sent to Segment from [[editedAutocompletedAddress]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "editedAutocompletedAddress",
+ *    context_module:"OrdersShipping",
+ *    context_page_owner_type: "orders-shipping",
+ *    context_page_owner_id: "57e60c68-a198-431e-8a02-6ecb01e3a99b",
+ *    field: "Address line 2"
+ *  }
+ * ```
+ */
+ export interface EditedAutocompletedAddress {
+  action: ActionType.editedAutocompletedAddress
+  context_module: ContextModule
+  context_owner_type: PageOwnerType
+  context_owner_id?: string
+  input: string
+  field: string
 }
--- a/src/Schema/Events/Search.ts
+++ b/src/Schema/Events/Search.ts
@@ -235,3 +235,77 @@ export interface SelectedSearchSuggestionQuickNavigationItem {
   destination_path: string
   label: "Artworks" | "Auction Results"
 }
+
+/**
+ * A User fills shipping address and has auto completion with results
+ *
+ * This schema describes events sent to Segment from [[addressAutoCompletionWithResults]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "addressAutoCompletionWithResults",
+ *    context_module:"OrdersShipping",
+ *    context_page_owner_type: "orders-shipping",
+ *    context_page_owner_id: "57e60c68-a198-431e-8a02-6ecb01e3a99b",
+ *    input: "Weserstr."
+ *  }
+ * ```
+ */
+ export interface AddressAutoCompletionWithResults {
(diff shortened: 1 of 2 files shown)


daytavares:
@erikdstock I am proposing three events:

- Address auto completion with results: the way it works for search is that we fire events each letter as the user inputs (see an example). Do you think it makes sense?
- Address auto completion with no results: do you think we need this one? Are there cases with no results?
- Selected address from the suggestions: I added a property that is "item" in case we can track the address selected.

I am assuming we don't need to add user_id here, but we will have this information sent.

Finally, in my notes I have: user sees a secondary suggestion. What are these cases?

Screenshot 2023-09-13 at 11 09 10

src/Schema/Events/Search.ts
```diff
@@ -235,3 +235,77 @@ export interface SelectedSearchSuggestionQuickNavigationItem {
   destination_path: string
   label: "Artworks" | "Auction Results"
 }
+
+/**
+ * A User fills shipping address and has auto completion with results
```
erikdstock, line 240: This will most likely appear on multiple pages, so we might want to just remove the specific reference to shipping here and rely on context module.

TMcMeans:
@daytavares can I ask a question about the proposal to track query events on each key press that the user inputs? I'm wondering how valuable that data is overall considering that there may be a lot of events coming through the pipeline. That may not be the case when we have requests throttled or debounced to limit the number of calls to the Smarty API.

I think it would be a good idea to track if the user doesn't select any addresses from the autocomplete dropdown? That maybe useful data for examining edge cases or the possibility that the user isn't receiving good query results.

erikdstock:
> @erikdstock I am proposing three events:
> 
> * Address auto completion with results: the way it works for search is that we fire events each letter as the user inputs (see an example). Do you think it makes sense?
> * Address auto completion with no results: do you think we need this one? Are there cases with no results?
> * Selected address from the suggestions: I added a property that is "item" in case we can track the address selected.
> 
> I am assuming we don't need to add user_id here, but we will have this information sent.
> 
> Finally, in my notes I have: user sees a secondary suggestion. What are these cases?
> 
> Screenshot 2023-09-13 at 11 09 10

Hi @daytavares and thanks- Overall this is and we could start from here. Biggest question below is #2. Looking at this together with @TMcMeans: 

1: What is the `id` in the screenshot? Do devs need to know about this?
2: How does data plan to use the events we send as the user types? We could be sending many events (every ~.5s as they type), so just want to make sure this is valuable on your side. So we could have a whole bunch of events with inputs like `123, 123 M, 123 Mai, 123 Main St`.
  - alternative here could be to send the # of fetches when the user ultimately selects an address (or doesn't?), but that might not work if they just never complete the form or exit early.
3: For `_WithResults` vs `_WithNoResults`: We could just include the count of results as part of the data for a single event.

daytavares:
Hi @erikdstock and @TMcMeans,

1. `id` in the screenshot is the event id
2. I tried to use the way we do for searches, but I am fine with # of fetches when the user ultimately selects an address (or doesn't?). I think it could be interesting to see # of fetches they need to enter to finally select the address (or not).
3. Yesss!

**Next steps (if you both agree):**
* Clean up `Address auto completion with results` and `Address auto completion with no results`
* Add properties # of fetches and # of results to `Selected address from the suggestions`

Do we need an event in case they don't select anything?

Thanks!

TMcMeans:
@daytavares I'm not sure but we may be able to track if the user clicks away from the dropdown menu instead of selecting a suggested option, but @erikdstock would know more about how feasible that will be to implement with the current UI setup.

daytavares:
@TMcMeans do you have any news here? :)

TMcMeans:
hey @daytavares I think it may be possible to track clicking away from the dropdown menu instead of selecting an option from the suggested addresses. Do you think that is good data to collect and analyze?

daytavares:
@TMcMeans yes! so I will create something similar to what we did for validation.

Do you agree with next steps?

- Clean up Address auto completion with results and Address auto completion with no results
- Add properties # of fetches and # of results to Selected address from the suggestions
- Create a `clicking away from the dropdown menu instead of selecting an option` event to track when users don't select any option

daytavares:
Hi @TMcMeans and @erikdstock,

I was thinking and @leamotta helped me with some ideas (thanks).

Basically what I want to know is: if the user inputs the address, I want to know if they selected or not a suggested address. 
Also interesting to know how many key presses and results we showed. 

So, two events:
* `FocusedOutAddresssBox`:  when an user focuses out an address line box (and there is input) -> here the idea is to know the users that used the address box
*  `SelectedItemFromAddressAutoCompletion`: user selects an item from address auto completion

With this I can calculate: from users who use the address box, how many select an option from auto completion. I think we can then see if the feature is useful or not.

erikdstock:
> Do we need an event in case they don't select anything? (I think so because I am wondering how I would know they don't select any option)

@daytavares - My thinking is that the first commit makes this possible - If a user sees autocomplete results, but they have no event for selecting an option, then they didn't select anything. Is it more complicated than that?

The approach in the current commit to track them clicking out of the box sounds harder to track for me - the user is always going to exit the input in some way, and not necessarily by clicking outside the suggestions. If i'm missing something about the importance of this event then we can go with what you have in 200e335, but I would suggest sticking with a schema like the first event like this:
```ts
 /* 
  Sent on every fetch + result for autocomplete suggestions - could be 5-10 per flow 
  An alternative would be to send only one event the first time they see suggestions and treat the event as a boolean.
*/
 export interface AddressAutoCompletionResult {
  action: ActionType.addressAutoCompletionWithResults
  context_module: ContextModule
  context_owner_type: PageOwnerType
  context_owner_id?: string
  input: string
  /* Count of results shown to user */
  count: number
}
```
- If we send an event for every new fetch of autocomplete results as above, we can see how many results were shown to the user on each one in case that is relevant (eg, did they eventually type so much that there were 0 results and proceed to enter a non-autocompleted address)
- If we only send a single event like 'yes, the user did see at least one autocomplete suggestion' then this would be a little bit simpler to implement, but not meaningfully. It comes down to what data would be useful for you.
- If we never send an event for the fetch, that would indicate the feature was not enabled.

One other thing we could track is whether the user edited the address _after_ selecting a suggestion. This could be an indication of how useful the suggestions were.
```ts
 /* MAYBE useful the first time they edit an autocompleted address */

 export interface EditedAutocompletedAddress {
  action: ActionType.addressAutoCompletionWithResults
  context_module: ContextModule
  context_owner_type: PageOwnerType
  context_owner_id?: string
  /* field they edited first (addressLine2 for example would suggest they are filling in their apt number) */
  field: string
}
```

src/Schema/Events/Search.ts
```diff
@@ -235,3 +235,79 @@ export interface SelectedSearchSuggestionQuickNavigationItem {
   destination_path: string
   label: "Artworks" | "Auction Results"
 }
+
+/**
+ * User fills shipping address and has/or not auto completion with results
```
erikdstock, line 240: nit: Could technically be any address in the future.

erikdstock (approved):
Great, thank you! I'll pull these requirements into our tracking ticket - should be fine as-is as long as you don't mind multiple tracking events as the results update.

artsyit:
:rocket: PR was released in [`v4.149.0`](https://github.com/artsy/cohesion/releases/tag/v4.149.0) :rocket:

PR #471, opened 2023-11-15
feat(ONYX-375): add ContextModule and OwnerType for alerts filters

dariakoko:

This PR resolves [ONYX-375]

### Description

Added ContextModule and OwnerType for alerts filters tracking on Eigen

[ONYX-375]: https://artsyproduct.atlassian.net/browse/ONYX-375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Schema/Values/ContextModule.ts
```diff
@@ -70,6 +70,7 @@ export enum ContextModule {
   curatedHighlightsRail = "curatedHighlightsRail",
   curatedTrendingArtistsRail = "curatedTrendingArtistsRail",
   curatorsPicksEmergingRail = "curatorsPicksEmergingRail",
+  currentAlertArtworkMatches = "currentAlertArtworkMatches",
```
dblandin, line 73: **suggestion**: Following up on [this Slack thread](https://artsy.slack.com/archives/C05EQL4R5N0/p1700049456288239), what do you think of `alertDetails`, `alertFilters`, and `alertConfirmation` as new context modules to add here?

src/Schema/Values/OwnerType.ts
```diff
@@ -6,6 +6,8 @@
 export enum OwnerType {
   activities = "activities",
   allArtistSeries = "allArtistSeries",
+  alertConfirmation = "alertConfirmation",
+  alertFilters = "alertFilters",
```
dblandin, line 10: **suggestion**: I don't think we need these values, which map to the specific steps in the alert flow. I think those are better captures as ContextModule values.

Although we might consider additions to eventually replace the existing `savedSearch` and `savedSearches` values such as `alert` and `alerts`.

dariakoko: Does that mean that you suggest excluding the owner_type from the event at all and use ContextModule instead? Wha value would you suggest to use for the ContextModule in this case?
[dock with the event details](https://docs.google.com/spreadsheets/d/1iBmzvJwgZ21J8w2VbVY3JAbH0EHbvUMz77HnZs3Ht9U/edit#gid=62721100)

dblandin: Ahh, you're totally right. For the app, we need these new owner types as there are distinct screens to track. In that case don't forget to include the in the `ScreenOwnerType` values below in this same file. While you're in here, maybe adding `alerts` and `alertDetails` as additional `ScreenOwnerType` values?

src/Schema/Values/ContextModule.ts
dariakoko: missed this one
adding now

artsyit:
:rocket: PR was released in [`v4.157.0`](https://github.com/artsy/cohesion/releases/tag/v4.157.0) :rocket:

PR #474, opened 2023-11-21
PR to update the version number

dariakoko:

This PR resolves [CO-]

### Description

Sorry for the mess!
I couldn't figure out how to upgrade the version number without using the Version label. I need some changes from the previous PRs. Those PRs had Version label Trivial which led to not changing the version number


artsyit:
:rocket: PR was released in [`v4.158.0`](https://github.com/artsy/cohesion/releases/tag/v4.158.0) :rocket:

PR #479, opened 2024-01-16
[AS-4346] Adding new events for partners offers

daytavares:

This PR resolves [AS-4346]

### Description
Creating events to track partners offers in CMS

[details](https://docs.google.com/spreadsheets/d/1c65vugkF2C4YuXwBXno8aqsBh4B2fTwxiLmgY_MhjWc/edit#gid=1515896528)

[AS-4346]: https://artsyproduct.atlassian.net/browse/AS-4346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -1873,36 +1873,40 @@ export interface ClickedAlertsFilters {
 /**
  * A user clicks on Start Sendind Offers on the marketing banner in CMS
  *
- * This schema describes events sent to Segment from [[ClickedStartSendingOffers]]
+ * This schema describes events sent to Segment from [[ClickedMarketingModal]]
  *
  *  @example
  *  ```
  *  {
- *    action: "clickedStartSendingOffers",
+ *    action: "clickedMarketingModal",
  *    context_page_owner_type: "Home",
  *    context_page_owner_id: "",
  *    context_page_owner_slug: "",
- *    partner_id: "61bcda16515b038ce5000104"
+ *    partner_id: "61bcda16515b038ce5000104",
+ *    user_id: "55bcda16515b038ce5033104",
+ *    title: "Make an offer, connect now"
  *  }
  * ```
  */
- export interface ClickedStartSendingOffers {
-  action: ActionType.clickedStartSendingOffers
+ export interface ClickedMarketingModal {
+  action: ActionType.clickedMarketingModal
   context_page_owner_type: PageOwnerType
   context_page_owner_id: string
   context_page_owner_slug?: string
   partner_id: string
+  user_id: string
+  title: string
 }
 
 /**
- * A user clicks on Send Offer on the Send Offers page in CMS
+ * A user clicks on the CTA to start with a partner offer on the Send Offers page in CMS
  *
- * This schema describes events sent to Segment from [[ClickedSendOffer]]
+ * This schema describes events sent to Segment from [[ClickedStartPartnerOffer]]
  *
  *  @example
  *  ```
  *  {
- *    action: "clickedSendOffer",
+ *    action: "clickedStartPartnerOffer",
  *    context_module: "Send offers",
  *    context_page_owner_type: "sendOffers",
  *    context_page_owner_id: "",
@@ -1916,8 +1920,8 @@ export interface ClickedAlertsFilters {
  *  }
  * ```
  */
- export interface ClickedSendOffer {
-  action: ActionType.clickedSendOffer
+ export interface ClickedStartPartnerOffer {
+  action: ActionType.clickedStartPartnerOffer
   context_page_module: string
   context_page_owner_type: PageOwnerType
   context_page_owner_id: string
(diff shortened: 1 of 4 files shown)


daytavares:
cc @artsy/emerald-devs

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -140,3 +140,94 @@ export interface ValidationAddressViewed {
...
+ *    partner_id: "61bcda16515b038ce5000104"
+ *  }
+ * ```
+ *
+ */
+export interface SendOffersBannerViewed {
+  action: ActionType.sendOffersBannerViewed
+  context_module: string
+  context_page_owner_type: PageOwnerType
+  partner_id: string
```
iskounen, line 164: This modal will be displayed to each browser that accesses the CMS, so it is expected that we have multiple `SendOffersBannerViewed` events for the same `partner_id`. Should we also track a `user_id` as well?

src/Schema/Events/Click.ts
```diff
@@ -1869,3 +1869,192 @@ export interface ClickedAlertsFilters {
...
+  artwork_id: string
+  price: number
+  last_offer_sent: number
+  new_saves: number
+  page: number
+}
+
+
+/**
+ * A user clicks on pagination on the Send Offers page in CMS
```
iskounen, line 1971: There are the arrows to paginate forward/backward, as well as the page numbers to jump to a specific page. Should this event be used for any type of pagination action?

src/Schema/Events/Click.ts
```diff
@@ -1869,3 +1869,192 @@ export interface ClickedAlertsFilters {
...
+  context_page_owner_id: string
+  context_page_owner_slug?: string
+  partner_id: string
+  artwork_id: string
+  price: number
+  collectors: number
+}
+
+/**
+ * A user clicks on Send Offers to Collectors in the modal for partner offers.
```
iskounen, line 2035: This description is a duplicate of the description for `ClickedSendOfferCollectors`. Which icon does this refer to?

iskounen (commented):
Unless I were very familiar with the current design of this feature in CMS, I would have a hard time understanding what these events are from just their names (the descriptions are helpful):

1. ClickedStartSendingOffers
1. ClickedSendOffer
1. ClickedSendOfferCollectors
1. ClickedOnSendOffersIcons

I worry that having event names that are so closely tied to the current UI will become a problem when the UI changes in V2. Would it be possible to rename them to match what the user is trying to achieve? For example:

1. Responded to marketing modal CTA
1. Started a partner offer
1. Sent a partner offer
1. (I'm not sure what this action is)

iskounen:
Should we also include an event to see if people are updating the artwork from the Send Offer modal?

Screenshot 2024-01-17 at 5 36 58 AM

src/Schema/Events/ImpressionTracking.ts
daytavares: Should I add user_id for all events? I thought since they are logged-in, user_id will be by default added.

src/Schema/Events/Click.ts
daytavares: Yes! That's why I added current and next page in case the label is arrow

src/Schema/Events/Click.ts
daytavares: I was thinking about our discussion if partners are clicking on ❤️ or the collector icon, but I substitute by ClickedUpdateArtwork (your next comment)

src/Schema/Events/ImpressionTracking.ts
iskounen: Oh! If that's already available, then please ignore this comment.

src/Schema/Events/Click.ts
rquartararo: Assuming we will use the `onClick` handler on `<ListPagniation>` to track this, I don't believe we will be able to distinguish if the click comes from the arrows or a page number (or which page number). The pagination component is in Palette which makes adding tracking more difficult. Would it be okay to only track the current page, and not the next page?

daytavares: Yes! I'll update here.

src/Schema/Events/Click.ts
iskounen: I remember our discussion about tracking clicks on row elements, but I've had doubts about how valuable that data would be, so I think this is a good change.

As a side note, one of the things we wanted to know is whether users would try to click on the artwork thumbnail or title to navigate to the artwork page. We've since decided to link those (and the artist name).

iskounen (approved):

artsyit:
:rocket: PR was released in [`v4.163.0`](https://github.com/artsy/cohesion/releases/tag/v4.163.0) :rocket:

PR #485, opened 2024-03-15
[AS-4428] Adding event to track list privacy

daytavares:

This PR resolves [[AS-4428](https://artsyproduct.atlassian.net/browse/AS-4428)]

### DIFF
--- a/src/Schema/Events/ArtworkLists.ts
+++ b/src/Schema/Events/ArtworkLists.ts
@@ -115,3 +115,25 @@ export interface ViewedArtworkList {
   context_owner_type: OwnerType.saves
   owner_id: string
 }
+
+/**
+ * User clicks on the toggle to edit artworks list privacy
+ *
+ * This schema describes events sent to Segment from [[editedArtworkListPrivacy]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "editedArtworkListPrivacy",
+ *    context_owner_type: "saves",
+ *    owner_id: "770fa47d-8cc8-4267-93e7-2808544d2a98",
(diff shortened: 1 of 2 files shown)


src/Schema/Events/ArtworkLists.ts
```diff
@@ -115,3 +115,25 @@ export interface ViewedArtworkList {
   context_owner_type: OwnerType.saves
   owner_id: string
 }
+
+/**
+ * User clicks on the toggle to edit artworks list privacy
+ *
+ * This schema describes events sent to Segment from [[editedArtworkListPrivacy]]
```
starsirius, line 122: @daytavares The [current design](https://www.figma.com/file/RHkhhJJwixhFPltqhH9ZkW/v1.5-%2B-v2-Saves-Partner-Offers?type=design&node-id=818-26825&mode=design&t=Zz8zXPyXWZaN8mBH-0) requires clicking the "Save Changes" button to save the list privacy update. It means:

- When collector toggles the list privacy for individual list, nothing is really saved, until clicking the "Save Changes" button. And it's possible that collector can toggle the privacy, dismiss the modal, and does _not_ save the changes.
- When clicking "Save Changes", we'll update the list privacy for _all_ lists in a batch.

So I want to double check that

- The event would only be sent when collector clicks "Save Changes", and would not be sent when collector toggles the privacy for individual list.
- We'll fire x number of events when collector saves the changes, with x being the total number of lists.
  - I wonder if it's better to just fire 1 event, with all the list privacy value in an array or something.

Let me know if I can pair and clarify things together!

daytavares:
@starsirius can we merge this?

PR #489, opened 2024-04-17
chore: AMBER-591 - add private artwork tracking methods

Mitch-Henson:

This PR partially resolves [AMBER-591]

### Description

This PR adds a few new events to deal with the new events that will take place on the newly created private artwork page.

[AMBER-591]: https://artsyproduct.atlassian.net/browse/AMBER-591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -2061,3 +2061,95 @@ export interface ClickedUpdateArtwork {
   partner_id: string
   artwork_id: string
 }
+
+/**
+ * A user clicks on the read more link on the private artwork page.
+ *
+ * This schema describes events sent to Segment from [[ClickedOnReadMore]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "clickedOnReadMore",
+ *    context_module: "About the work" | "About the artist" | "Sidebar",
+ *    subject: "Read more"
+ *    type: "Link"
+ *  }
+ * ```
+ */
+export interface ClickedOnReadMore {
+  action: ActionType.clickedOnReadMore
+  context_module: string
+  subject: string
+  type: string
+}
+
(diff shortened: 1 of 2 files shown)


jpotts244 (approved):
Looks good on my end! Would love to hear @louislecluse 's thoughts

damassi (approved):

damassi:
Looks good!

src/Schema/Events/Click.ts
```diff
@@ -2061,3 +2061,95 @@ export interface ClickedUpdateArtwork {
...
+ *  {
+ *    action: "clickedOnGalleryName",
+ *    context_module: "Sidebar",
+ *    subject: "Gallery Name"
+ *    type: "Link"
+ *    flow: "Exclusive Access"
+ *  }
+ * ```
+ */
+export interface ClickedOnGalleryName {
```
louislecluse, line 2127: We can just use a generic click event for this one! It sounds a bit too specific to track click gallery name

Mitch-Henson: Ok great! Removed this in the [newest PR](https://github.com/artsy/cohesion/pull/490). I also wanted to ask about whether we want to track clicking on the artist name like we do with the gallery name? That would probably also use the generic click event as well then, right?

The image below highlights what we are going to track with this 591 ticket, with the exception of the artist.

Screenshot 2024-04-19 at 13 54 46

louislecluse: The artist name is not specific to the private artwork, so don't need to track for now!

PR #497, opened 2024-05-06
fix: broken TappedBuyNow flow type

MounirDhahri:

### Description

Hey @daytavares - I noticed that this change broke the type check CI step on force https://github.com/artsy/force/pull/13792. I believe some changes need to be made there in order to support the new property "flow" since it's now required. 

Because I don't see any PR doing that currently, I am adjusting the type, temporarily hopefully until such changes are made in Force, to unblock the cohesion upgrade. 

I also noticed that the `flow` property is a string, I changed that to a literal string to make sure we don't send a wrong string.

### DIFF
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -798,7 +798,7 @@ export interface TappedBuyNow {
   context_owner_id: string
   context_owner_slug: string
   impulse_conversation_id?: string
-  flow: string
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in [`v4.179.0`](https://github.com/artsy/cohesion/releases/tag/v4.179.0) :rocket:

daytavares:
Thanks!!

PR #498, opened 2024-05-06
chore: Adds optional flow metadata to TappedLearnMore interface

jpotts244:
The type of this PR is: Chore

### Description

Noticed a discrepancy issue between the `Click*` and `Tap*` events while working on the Private Artworks feature in eigen
https://github.com/artsy/eigen/pull/10195

From this project's tracking requirements, we do want to track additional `flow` metadata attrs on the `TappedLearnMore` type events so adding a new optional field to `TappedLearnMore`

Without adjusting this type, we are doing a lot of events by hand in eigen


louislecluse (approved):

damassi (approved):

artsyit:
:rocket: PR was released in [`v4.180.0`](https://github.com/artsy/cohesion/releases/tag/v4.180.0) :rocket:

PR #500, opened 2024-05-08
fix: updates the context_owner_type for ClickedBuyNow event

MrSltun:

This PR resolves []

### Description
This PR updates the type of `context_owner_type` in the `ClickedBuyNow` event to accept both Notification and Artwork pages, similar to `TappedBuyNow`


leamotta (approved):

artsyit:
:rocket: PR was released in [`v4.182.0`](https://github.com/artsy/cohesion/releases/tag/v4.182.0) :rocket:

PR #509, opened 2024-06-28
amending CollectorProfiles for the new flow

leodmz:
The values for `ContextModule` and `ContextScreenOwnerType` were hard coded in the Collector profile event. 
With the new flow we're adding this event on other surfaces and this change will allow us to track that


iskounen (approved):
Thanks Léo 🎉

artsyit:
:rocket: PR was released in [`v4.190.0`](https://github.com/artsy/cohesion/releases/tag/v4.190.0) :rocket:

PR #510, opened 2024-07-02
feat: Adding tracking for User prompting flow

leodmz:
[here](https://www.notion.so/artsy/Second-launch-tracking-and-impact-e5a4199344424c9e98a2d973183af4cf) is the tracking plan

for the different flows we want to track
1. when a user entered the flow (either with an impression tracking or a click)
2. when a user finished the flow by clicking save and exit

We cannot have just 2 events because the flows described [here](https://www.figma.com/design/yugauriWRATcpu6knETvRY/Collector-Profile---User-Facing?node-id=1702-17470&t=g0QCAtLx2xbpmsnp-0) and [here](https://www.figma.com/design/y68lismoyIR69mTYtEcmrD/Collector-Prompting?node-id=1-324&t=vmgsKZwz2iiBRec1-0) behave differently

This PR adds: 
* `editprofileModalviewed` event to track when the modal was displayed after an inquiry
* capabilities to track if the event was fired on app or web in `EditedUserProfile` and `TappedMyCollectionAddArtworkArtist`

### DIFF
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -1,6 +1,7 @@
 import { ContextModule } from "../Values/ContextModule"
 import { OwnerType, PageOwnerType } from "../Values/OwnerType"
 import { ActionType } from "."
+import { Platform } from "./MyCollection"
 
 /**
  * Schemas describing rail_viewed events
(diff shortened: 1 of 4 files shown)


src/Schema/Events/ImpressionTracking.ts
```diff
@@ -231,3 +232,28 @@ export interface SendOffersErrorMessage {
...
+ * This schema describes events sent to Segment from [[EditProfileModalViewed]].
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "editProfileModalViewed",
+ *    context_module: "add artwork" or "add profession/location"
+ *    context_page_owner_type: "home"
+ *    partner_id: "61bcda16515b038ce5000104"
+ *  }
```
MounirDhahri, line 248:
```suggestion
 *    partner_id: "61bcda16515b038ce5000104"
 *    platform: "web"
 *    user_id: "61bcda16515b038ce5000104"
 *  }
```

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -231,3 +232,28 @@ export interface SendOffersErrorMessage {
...
+ * ```
+ *
+ */
+export interface EditProfileModalViewed {
```
MounirDhahri, line 252: You might need to export this `Events/index.ts`

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -231,3 +232,30 @@ export interface SendOffersErrorMessage {
   collectors: number
   message: string
 }
+
+/**
+ * User sees the edit profile modal after sending an inquiry
```
iskounen, line 237: The user could also be shown the "add artist to my collection" modal (and close it without adding an artists) after sending an inquiry, so we'll need another impression event for that.

artsyit:
:rocket: PR was released in [`v4.191.0`](https://github.com/artsy/cohesion/releases/tag/v4.191.0) :rocket:

PR #512, opened 2024-07-09
feat: Adding events for App upsell tracking

leodmz:
adding two events for the app upsell tracking 
both events are click events fired when a user clicks on either of the modals taking them to the app store 
the first one is the header present on all mweb pages and the second one is the footer present after the second page view.

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -2191,3 +2191,33 @@ export interface ClickedCompleteYourProfile {
   context_page_owner_slug?: string
   user_id: string
 }
+
+/**
+ * A user clicks on the header taking them to the app store 
+ *
+ * This schema describes events sent to Segment from [[clickedDownloadAppHeader]]
+ *
+ */
+
+export interface ClickedDownloadAppHeader {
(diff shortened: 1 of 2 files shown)


artsyit:
:rocket: PR was released in [`v4.193.0`](https://github.com/artsy/cohesion/releases/tag/v4.193.0) :rocket:

src/Schema/Events/Click.ts
```diff
@@ -2190,4 +2190,34 @@ export interface ClickedCompleteYourProfile {
...
+ * This schema describes events sent to Segment from [[clickedDownloadAppHeader]]
+ *
+ */
+
+export interface ClickedDownloadAppHeader {
+  action: ActionType.clickedDownloadAppHeader
+  context_page_owner_type: PageOwnerType
+  context_page_owner_id: string
+  context_page_owner_slug?: string
+  user_id: string
```
egdbear, line 2207: @leodmz 

I think we should change this (also ClickedDownloadAppFooter) to:  

```
  context_page_owner_slug?: string
  user_id?: string 
 ```
This should be optional because the app download banner is shown also to non-logged-in users, right? 

cc @dzucconi

PR #518, opened 2024-08-06
fix: updating the app upsell tracking events

leodmz:
Adding user_id and context_page_owner_id as optional to make sure we're able to track clicks from logged out users


egdbear (approved):
✅

artsyit:
:rocket: PR was released in [`v4.198.0`](https://github.com/artsy/cohesion/releases/tag/v4.198.0) :rocket:

PR #523, opened 2024-08-27
feat: adding a value for the new first step of the signup flow

leodmz:
The new sign up/log in flow introduces a first step where the user has to fill out their email address first to then be automatically redirected to either the sign up or log in flow. 
This new value allows us to track when that first modal is displayed (called "welcome")

[Figma designs](https://www.figma.com/design/jFcE32eMExDPj0XgmIThML/Sign-up---Improvements?node-id=54-2770&t=aedNNnadwG5wJBlF-0)

### DIFF
--- a/src/Schema/Events/Authentication.ts
+++ b/src/Schema/Events/Authentication.ts
@@ -149,6 +149,8 @@ export enum AuthModalType {
   login = "login",
   /** Sign up for a new account */
   signup = "signup",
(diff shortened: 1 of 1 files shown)


dzucconi:
Thanks! Need a new intent as well: https://github.com/artsy/cohesion/blob/e12f4b5beebf13d96c60b6ff0905ef85da10acc9/src/Schema/Values/Intent.ts#L7

PR #526, opened 2024-09-12
feat: added tappedNewsGroup

MounirDhahri:

This PR resolves [ONYX-1288]

### Description

This PR adds `TappedNewsGroup` as an event. This is required for the news section in Home. 

**Note:** 
- In the existing home view, we are not tracking taps on the news articles.
- I did some alphabetical sorting to make additions easier in the past

[ONYX-1288]: https://artsyproduct.atlassian.net/browse/ONYX-1288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -44,12 +44,11 @@ export interface TappedArticleGroup extends TappedEntityGroup {
  *    action: "tappedNewsGroup",
  *    context_module: "articleRail",
  *    context_screen_owner_type: "home",
- *    context_screen_owner_id: "5e726bd22524980012caafb0",
- *    context_screen_owner_slug: "arteba-special-edition",
  *    destination_screen_owner_type: "article",
  *    destination_screen_owner_id: "542f1ccc7261694847410400",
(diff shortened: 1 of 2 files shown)


src/Schema/Events/Tap.ts
```diff
@@ -33,6 +33,29 @@ export interface TappedArticleGroup extends TappedEntityGroup {
...
 
+/**
+ * A user taps a grouping of news on App
+ *
+ *  This schema describes events sent to Segment from [[tappedNewsGroup]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "tappedNewsGroup",
```
dblandin, line 44: **question**: Don't we already have a `tappedArticleGroup` event we could use?

MounirDhahri: hmm, that's true - it seems like news are articles 😄

MounirDhahri: I will update the payload to reflect the type of the section (rail / list)

PR #527, opened 2024-09-18
chore: add `marketNews` as PageOwner type (page+screen)

dblandin:

### Description

To differentiate between the "Market News" landing page and "Artsy Editorial".


artsyit:
:rocket: PR was released in [`v4.205.0`](https://github.com/artsy/cohesion/releases/tag/v4.205.0) :rocket:

PR #531, opened 2024-10-11
chore: Add more data to event when publishing matching work from demand AMBER-1090

jonallured:
This PR updates our event for when a partner publishes a matching artwork in CMS. It now includes a the `PartnerSearchCriteria` id and the page rank of the alert that's being published against.

https://artsyproduct.atlassian.net/browse/AMBER-1090

/cc @artsy/amber-devs @leodmz

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -1812,21 +1812,25 @@ export interface ClickedNavBar {
  *  ```
  *  {
  *    action: "clickedUploadArtwork",
- *    context_module: "alerts-price"
- *    context_page_owner_type: "demand",
- *    context_page_owner_id: "",
+ *    alert_page_rank: 1,
  *    artist_id: "4212691337420",
- *    search_criteria_id: "4212691337420"
+ *    context_module: "alerts-price",
+ *    context_page_owner_id: "",
+ *    context_page_owner_type: "demand",
+ *    partner_search_criteria_id: "123abc",
+ *    search_criteria_id: "4212691337420",
  *    subject: "Upload Artworks"
  *  }
  * ```
  */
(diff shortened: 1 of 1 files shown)


src/Schema/Events/Click.ts
```diff
@@ -1812,21 +1812,25 @@ export interface ClickedNavBar {
...
+  alert_page_rank?: number
+  artist_id?: string
   context_module: ContextModule
-  context_page_owner_type: PageOwnerType
   context_page_owner_id?: string
-  artist_id?: string
+  context_page_owner_type: PageOwnerType
+  partner_search_criteria_id?: string
   search_criteria_id?: string
   subject: string
```
jpotts244, line 1835: Just to clarify for me, we do still need both `partner_search_criteria_id` and `search_criteria_id`?

jonallured: Technically we probably don't need `search_criteria_id` but I don't think it hurts anything. It's also the case that past events will have this id in segment so idk if there are implications there? /cc @leodmz

jpotts244: 🙏🏻 up to you guys!

jpotts244 (approved):
LGTM

src/Schema/Events/Click.ts
leodmz: coming back to this, if past events have this info I think we have to keep it to avoid breaking anything downstream

jonallured:
Merging!

artsyit:
:rocket: PR was released in [`v4.209.0`](https://github.com/artsy/cohesion/releases/tag/v4.209.0) :rocket:

PR #533, opened 2024-10-23
feat(EMI-2098): Add tappedNotification and tappedClearNotification to the schema

MrSltun:

This PR resolves [EMI-2098]

### Description
This PR adds `TappedNotification` and `TappedClearNotification` events to `Tap.ts`

[EMI-2098]: https://artsyproduct.atlassian.net/browse/EMI-2098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -244,7 +244,6 @@ export interface TappedBrowseSimilarArtworks extends TappedEntityGroup {
  *    action: "tappedClearNotification",
  *    context_module : "Home",
  *    context_screen_owner_type: "Home",
- *    context_screen_owner_id: "58de681f275b2464fcdde097",
  *    notification_id: "23424132",
  *    notification_category: "send_wire"
  *  }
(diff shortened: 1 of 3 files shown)


MrSltun (commented):
@daytavares I noticed that we already have `notification` in [ContextModule](https://github.com/artsy/cohesion/blob/main/src/Schema/Values/ContextModule.ts#L143), should we use it or create a new one names `notifications` or `tasks`?

daytavares:
> @daytavares I noticed that we already have `notification` in [ContextModule](https://github.com/artsy/cohesion/blob/main/src/Schema/Values/ContextModule.ts#L143), should we use it or create a new one names `notifications` or `tasks`?

I don't know where `notification` is used, but I guess it's good if we are consistent across railViewed and tapped events. Can we create `actNow`. I guess we ended up with this name for the zone, right?

daytavares (approved):
LGTM! Thanks!

PR #534, opened 2024-10-23
feat: update events definition to track update failed payments

oxaudo:

This PR adds events to resolve [[EMI-2130](https://artsyproduct.atlassian.net/browse/EMI-2130)] and [[EMI-2131](https://artsyproduct.atlassian.net/browse/EMI-2131)]

### Description

@daytavares and myself aligned that `clickedChangePaymentMethod` with proper `context_module` and `context_page_owner_type` should be enough to track clicks to update failed payment on the web.

For the app we are adding 2 events `bannerViewed` to track impressions for the banner on the home screen and `tappedChangePaymentMethod` to track request to update payment method from both paymentFailed banner and orders history screen.

- [X] If I've added a new file to the tree I've exported it from the common [`index.ts`](https://github.com/artsy/cohesion/blob/main/src/Schema/Events/index.ts)
- [X] I've added comments with examples for any new interfaces and ensured that they're in the docs
- [X] No platform-specific terminology has been used outside of `click` and `tap` (platform is inferred by the DB storing events)

[EMI-2130]: https://artsyproduct.atlassian.net/browse/EMI-2130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMI-2131]: https://artsyproduct.atlassian.net/browse/EMI-2131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Schema/Events/Click.ts
```diff
@@ -373,8 +373,8 @@ export interface ClickedBid {
  *  ```
  *  {
  *    action: "clickedChangePaymentMethod",
- *    context_module: "OrdersReview",
- *    context_page_owner_type: "orders-review",
+ *    context_module: "OrdersReview" | "OrdersHistory",
```
oxaudo, line 376: Added this as an example of how we plan to use this on the orders history page. Think we'll put specific `order_id` from the row that displays update `update payment` button  into `context_page_owner_id` - which is theoretically not correct as `ordersHistory` page does not include id. But this id will point to the corresponding row. You think it's acceptable @daytavares ? Alternatively we can add `item_id` to this schema to be more proper with our assignment.

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -259,3 +259,28 @@ export interface EditProfileModalViewed {
...
+ * User sees the banner in the app
+ *
+ * This schema describes events sent to Segment from [[BannerViewed]].
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "bannerViewed",
+ *    context_screen: "home",
+ *    context_module: "paymentFailed",
```
oxaudo, line 273: I added this `paymentFailed` as a `context_module` which in combination with `bannerViewed` and `context_screen` will allow to indicate `paymentFailed banner on home screen`. Think it's descriptive enough. If we potentially put this same banner on other screens we'll be able to call the same with different `context_screen`. That makes sense to me.

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -259,3 +259,28 @@ export interface EditProfileModalViewed {
...
+ *    item_type: "order" | "orders"
```
oxaudo, line 274: This will allow us to differentiate if the banner was for one failed payment or multiple orders with failed payments. Does that make sense? Is that how we can do this? Is it important to track if people see banner for one failed payment or multiple failed payments?

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -259,3 +259,28 @@ export interface EditProfileModalViewed {
...
+ *    item_id: "order-id" | ""
```
oxaudo, line 275: And this `item-id` is order id with failed payment in case of a single failing payment or blank if there were multiple failed payments and we are just linking to orders history page. Makes sense?

src/Schema/Events/Tap.ts
```diff
@@ -1176,3 +1176,28 @@ export interface TappedCompleteYourProfile {
...
+ *
+ * This schema describes event sent to Segment from [[tappedChangePaymentMethod]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "tappedChangePaymentMethod",
+ *    context_module : "ordersHistory" | "paymentFailed"
+ *    context_screen "orders-history" | "home",
+ *    item_id: "order-id" | ""
```
oxaudo, line 1191: same logic for one failed payment / multiple payments failed described above.

src/Schema/Events/Tap.ts
```diff
@@ -1176,3 +1176,28 @@ export interface TappedCompleteYourProfile {
...
+/**
+ * A user taps to request a chage of payment method for the order/orders
+ *
+ * This schema describes event sent to Segment from [[tappedChangePaymentMethod]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "tappedChangePaymentMethod",
+ *    context_module : "ordersHistory" | "paymentFailed"
```
oxaudo, line 1189: and in here we are capturing if the button was in paymentFailed banner or on the `ordersHistory` page. Is that understandable distinction? Proper way to encode those options?

oxaudo (commented):
Tried to summarize my thinking of usage inline. Please let me know if this sounds legit.

src/Schema/Events/Click.ts
daytavares: I think this is good, so we can track the `order_id`.

src/Schema/Events/ImpressionTracking.ts
daytavares: I think it's not harm, but also not super necessary. The idea is understanding more generally if the view the banner > clicked on it

oxaudo:
Ok. Seems like no changes are requested here. Merging this so we can use this in the code.

artsyit:
:rocket: PR was released in [`v4.212.0`](https://github.com/artsy/cohesion/releases/tag/v4.212.0) :rocket:

src/Schema/Events/Tap.ts
```diff
@@ -1228,3 +1228,28 @@ export interface TappedCompleteYourProfile {
...
+ *    item_type: "order" | "orders"
+ *  }
+ * ```
+ */
+
+export interface TappedChangePaymentMethod {
+  action: ActionType.tappedChangePaymentMethod
+  context_module: ContextModule
+  context_screen: OwnerType
+  idem_id: string
```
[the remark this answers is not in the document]
oxaudo: Oh my! Thanks for noticing! On it:)

PR #535, opened 2024-10-24
feat(DIA-863): add discoverSomethingNewSection and exploreBySection modules

araujobarret:
The type of this PR is: **Feature**

This PR resolves [DIA-863]

### Description

Adds the Context Modules `discoverSomethingNewSection` and `exploreBySection`.

### PR Checklist (tick all before merging)

- [x] If I've added a new file to the tree I've exported it from the common [`index.ts`](https://github.com/artsy/cohesion/blob/main/src/Schema/Events/index.ts)
- [x] I've added comments with examples for any new interfaces and ensured that they're in the docs
- [x] No platform-specific terminology has been used outside of `click` and `tap` (platform is inferred by the DB storing events)

[DIA-863]: https://artsyproduct.atlassian.net/browse/DIA-863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #536, opened 2024-10-24
feat(DIA-863): add tap events to the new home view cards

araujobarret:

This PR resolves [DIA-863]

### Description

Sibling of #535

Adds:
- `TappedCardsGroup` and `TappedCardsChipsGroup` events
- `collectionsCategory` as owner type

[DIA-863]: https://artsyproduct.atlassian.net/browse/DIA-863?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Schema/Events/Tap.ts
```diff
@@ -327,6 +327,50 @@ export interface TappedFairGroup extends TappedEntityGroup {
...
+ *    action: "tappedCardsGroup",
+ *    context_module: "exploreBySection",
+ *    context_screen_owner_type: "home",
+ *    destination_screen_owner_type: "collectionsCategory",
+ *    destination_screen_owner_slug: "alex-katz-black-dress",
+ *    type: "thumbnail"
+ *  }
+ * ```
+ */
+export interface TappedCardsGroup extends TappedEntityGroup {
```
dblandin, line 347: We usually have the entity in the event name singular: so `tappedCardGroup`.

src/Schema/Events/Tap.ts
```diff
@@ -327,6 +327,50 @@ export interface TappedFairGroup extends TappedEntityGroup {
...
+ *    context_screen_owner_type: "home",
+ *    destination_screen_owner_type: "collection",
+ *    destination_screen_owner_id: "5359794d1a1e86c3740001f7",
+ *    destination_screen_owner_slug: "alex-katz-black-dress",
+ *    horizontal_slide_position: 1,
+ *    type: "thumbnail"
+ *  }
+ * ```
+ */
+export interface TappedCardsChipsGroup extends TappedEntityGroup {
```
dblandin, line 370: Still not sold on needing both these events. Maybe just start with `tappedCardGroup`? We'll be able to use the contextModule value to differentiate between specific section instances.

egdbear (approved):

dblandin (commented):
Looking good. Left a couple comments but let's see what @louislecluse thinks.

src/Schema/Events/Tap.ts
araujobarret: makes sense, removed in the last commits 👍

src/Schema/Events/Tap.ts
araujobarret: changed it to singular in the latest commits 👍

louislecluse (approved):
@daytavares We're introducing "tappedCardGroup" here (for a new component type closer to a 'card'), I think it works with your tracking guideline but let me know otherwise.

daytavares (approved):

src/Schema/Events/Tap.ts
```diff
@@ -327,6 +327,29 @@ export interface TappedFairGroup extends TappedEntityGroup {
   action: ActionType.tappedFairGroup
 }
 
+/**
+ * A user taps a grouped card (Explore By, Discover Something new)
```
dblandin, line 331: **suggestion**: "A user taps a grouping of cards (one of our supported home view section types)"? Maybe drop the specific "Explore By" / "Discover Something New" examples. Not sure if that's super meaningful out of context.

src/Schema/Events/Tap.ts
```diff
@@ -327,6 +327,29 @@ export interface TappedFairGroup extends TappedEntityGroup {
...
+ *
+ * This schema describes events sent to Segment from [[tappedEntityGroup]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "tappedCardGroup",
+ *    context_module: "exploreBySection" | discoverSomethingNewSection,
```
dblandin, line 339: **suggestion**: I'd pick one of these strings for the example and drop the other as it becomes invalid JSON.

src/Schema/Events/Tap.ts
```diff
@@ -327,6 +327,29 @@ export interface TappedFairGroup extends TappedEntityGroup {
...
+ *    context_screen_owner_type: "home",
+ *    destination_screen_owner_type: "collectionsCategory" | "collection",
```
dblandin, line 341: **suggestion**: Same here, I'd pick one string to use in the example.

src/Schema/Events/Tap.ts
```diff
@@ -327,6 +327,29 @@ export interface TappedFairGroup extends TappedEntityGroup {
...
+ *    destination_screen_owner_id: "5359794d1a1e86c3740001f7" | undefined,
```
dblandin, line 342: **suggestion**: Even though technically this can be undefined as the type allows, I'd specify just the string value here in the example.

src/Schema/Events/Tap.ts
```diff
@@ -327,6 +327,29 @@ export interface TappedFairGroup extends TappedEntityGroup {
...
+ *    destination_screen_owner_slug: "artworks-under-1000",
+ *    horizontal_slide_position: 1 | undefined,
```
dblandin, line 344: **suggestion**: Same here. Even though technically this can be undefined as the type allows, I'd specify just the integer value here in the example.

dblandin (approved):
Awesome! Just a few minor suggestions for description/example in the code comment.

araujobarret:
Thanks for the detailed review Devon, I've addressed the last comments here https://github.com/artsy/cohesion/pull/536/commits/21f319ec0b4d5b6a16a630e06969714e2ca455fa and renamed the respective context modules to match what we agreed upon 👍 
once that's 🟢 and merged will update the MP PR

src/Schema/Values/OwnerType.ts
```diff
@@ -33,6 +33,7 @@ export enum OwnerType {
   collect = "collect",
   collection = "collection",
   collections = "collections",
+  collectionsCategory = "collectionsCategory",
```
dblandin, line 36: This is very easy to miss (and annoying to discover later) but this value won't be easily accessible by Eigen unless it's also included in the ScreenOwnerType definition below ~L158.

araujobarret: oh good catch, I think we can enhance these typings(ScreenOwnerType) in the future with Omit and Pick to make it lighter for engineers working with it

dblandin: Yeah, for engineers that truly want to work around the typing they can always force it with `"somethingElse" as ScreenOwnerType` but at least that's made explicit in the downstream code.

artsyit:
:rocket: PR was released in [`v4.213.0`](https://github.com/artsy/cohesion/releases/tag/v4.213.0) :rocket:

PR #538, opened 2024-11-06
feat: add tappedTaskGroup and tappedClearTask

xander-pero:

This PR resolves [EMI-2167] and [AS-4701]

### Description

Renaming the tappedNotification to tappedTaskGroup and tappedClearNotification to tappedClearTask (to match former naming). This matches the naming schema on the homeview; for the Emerald negotiation zone. 

[EMI-2167]: https://artsyproduct.atlassian.net/browse/EMI-2167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AS-4701]: https://artsyproduct.atlassian.net/browse/AS-4701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -372,6 +372,7 @@ export interface TappedEntityGroup {
     | ActionType.tappedFairGroup
     | ActionType.tappedViewingRoomGroup
     | ActionType.tappedHeroUnitGroup
+    | ActionType.tappedTaskGroup
   context_module: ContextModule
   context_screen_owner_type: ScreenOwnerType
   context_screen_owner_id?: string
@@ -1188,12 +1189,12 @@ export interface TappedNewsSection {
 /**
  * A user taps on a task notification in the app
  *
- * This schema describes events sent to Segment from [[tappedNotification]]
+ * This schema describes events sent to Segment from [[tappedTaskGroup]]
  *
  *  @example
  *  ```
  *  {
- *    action: "tappedNotification",
+ *    action: "tappedTaskGroup",
  *    context_module : "Home",
  *    context_screen_owner_type: "Home",
  *    destination_path: "/orders/123",
@@ -1202,10 +1203,8 @@ export interface TappedNewsSection {
  *  }
  * ```
  */
-export interface TappedNotification {
-  action: ActionType.tappedNotification
-  context_module: ContextModule
-  context_screen_owner_type: ScreenOwnerType
+export interface TappedTaskGroup extends TappedEntityGroup {
+  action: ActionType.tappedTaskGroup
   destination_path: string
   notification_id: string
   notification_category: string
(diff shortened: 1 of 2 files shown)


src/Schema/Events/Tap.ts
```diff
@@ -1202,10 +1203,8 @@ export interface TappedNewsSection {
...
  */
-export interface TappedNotification {
-  action: ActionType.tappedNotification
-  context_module: ContextModule
-  context_screen_owner_type: ScreenOwnerType
+export interface TappedTaskGroup extends TappedEntityGroup {
+  action: ActionType.tappedTaskGroup
   destination_path: string
   notification_id: string
   notification_category: string
```
dblandin, line 1210: Should we also rename these payload keys to `task_[field]`? Is a task's `category` expected to be the same value as the task's `type` in Gravity?

I don't see category in the backend data modeling but a [task's type](https://github.com/artsy/gravity/blob/95e9760d3512a8b8ad1d250c6b7d141938579196/app/models/domain/task.rb#L2) can currently be one of send_wire, outbid, counteroffer, or id_verification.

xander-pero: I agree with the change to rename to `task_[field]`, overlooked that when fixing the naming.

For your second question, do you suggest renaming the (now) `task_category` into `task_type`, or to deprecate `task_category` and instead make use of `type` inherited from TappedEntityGroup? (Would the TappedEntityGroup type link to the `TYPES` referenced in [task.rb](https://github.com/artsy/gravity/blob/95e9760d3512a8b8ad1d250c6b7d141938579196/app/models/domain/task.rb)?

dblandin: I think it's fine to extend the TappedEntityGroup interface with two new custom fields to track task metadata (id and type). I wouldn't reuse the existing `type` (not prefixed) property on TappedEntityGroup as that refers to something else.

As far as I can tell, the Task model defined in Gravity doesn't yet have a "category" concept. If we're instead expecting the internal "task type" value here, we can use a similar property name to avoid conflicts in the future if we ever want to introduce categories to the Task model.

So the full interface might look something like this:

```ts
export interface TappedTaskGroup extends TappedEntityGroup {
  action: ActionType.tappedTaskGroup
  destination_path: string
  task_id: string
  task_type: string
Member

```

xander-pero: I see, this makes sense. I misunderstood what you were trying to communicate before, and agree with this change. Thank you for your help and the completed example!

xander-pero:
Renamed `notification_id` to `task_id` and `notification_category` to `task_type`

MrSltun (approved):
Looks great! Thank you Xander 🌟

dblandin (approved):
Awesome. Thanks @xander-pero!

artsyit:
:rocket: PR was released in [`v4.215.0`](https://github.com/artsy/cohesion/releases/tag/v4.215.0) :rocket:

PR #548, opened 2025-01-16
chore: rename quickLinks to navigationPills

dariakoko:

This PR resolves [CO-]

### Description

As Recommended [here](https://github.com/artsy/metaphysics/pull/6354#pullrequestreview-2555968879) rename quckLinks to navigationPills

### DIFF
--- a/src/Schema/Values/ContextModule.ts
+++ b/src/Schema/Values/ContextModule.ts
@@ -145,6 +145,7 @@ export enum ContextModule {
   myCollectionMarketSignals = "myCollectionMarketSignals",
   myCollectionOnboarding = "myCollectionOnboarding",
   navBar = "navBar",
+  navigationPills = "navigationPills",
   newWorksByArtistsYouFollowRail = "newWorksByArtistsYouFollowRail",
   newWorksByGalleriesYouFollowRail = "newWorksByGalleriesYouFollowRail",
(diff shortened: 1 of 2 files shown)


MounirDhahri (changes requested):
I believe we can keep on calling this quick pills but only have the section type to be different.

dariakoko:
I think it's best to be consistent with naming. Is there a reason you want to call it different names?

dblandin:
> I think it's best to be consistent with naming. Is there a reason you want to call it different names?

@dariakoko In this case the difference in naming is intentional. The schema design allows for a _section type_ to be reused by multiple _section instances_. For example, we have the following relationships already

- Artworks > NewWorksForYou, RecommendedArtworks, SimilarToRecentlyViewedArtworks, AuctionLotsForYou, ActiveBids
- Artists > RecommendedArtists, TrendingArtists
- MarketingCollections > CuratorsPicksEmerging
- HomeViewSectionCards > ExploreByCategory, DiscoverSomethingNew

So keeping the names difference would introduce a similar relationship:

- NavigationPills > QuickLinks, [...FutureSection...]

dblandin:
Regarding analytics, we should refer to the section instance's name – not the section types so that we're specific with that we're referring to.

PR #553, opened 2025-02-03
chore: Add click event for saving an artwork in CMS edit form AMBER-1268

jonallured:
There are two buttons in CMS for the Artwork Form:

sabretooth-000052@2x

Neither of these buttons are currently being tracked but I'm working on fixing this! When I came to Cohesion to find a suitable event I found `ClickedPublish`:

```ts
/**
 * A partner clicks the publish button on the artwork form page in CMS.
 *
 * This schema describes events sent to Segment from [[ClickedPublish]]
 *
 * @example
 * ```
 * {
 *    action: "clickedPublish",
 *    context_module: "artworkDetails" ,
 *    artwork_id: "60de173a47476c000fd5c4cc"
 *    label: "Publish"
 * }
 * ```
 */

export interface ClickedPublish {
  action: ActionType.clickedPublish
  context_module: ContextModule
  artwork_id: string
  label: string
}
```

So then I thought I'd add the Save version and that's what's in this PR. This may be good enough but wait a minute - if the publish button in the artwork form is not currently being tracked then how are we even using `ClickedPublish`?

Turns out we use this event on the Artwork show page! Here:

sabretooth-000053@2x

Which.....is not what the description of the event says - that note says that this is used on the artwork form. Is this a problem? I guess we could use the `context_module` to disambiguate. So that's what I did with the second commit - added the `ContextModule.artworkForm` option. Am I missing something or is this cool?

https://artsyproduct.atlassian.net/browse/AMBER-1268

/cc @artsy/amber-devs

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -1555,6 +1555,29 @@ export interface ClickedPublish {
 }
 
 /**
+ * A partner clicks the save button on the artwork form page in CMS.
+ *
+ * This schema describes events sent to Segment from [[ClickedSave]]
+ *
+ * @example
+ * ```
+ * {
+ *    action: "clickedSave",
+ *    context_module: "artworkDetails" ,
+ *    artwork_id: "60de173a47476c000fd5c4cc"
+ *    label: "Save"
+ * }
+ * ```
+ */
+
+export interface ClickedSave {
+  action: ActionType.clickedSave
+  context_module: ContextModule
+  artwork_id: string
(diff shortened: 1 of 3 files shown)


damassi (approved):

damassi:
All makes sense 👍

artsyit:
:rocket: PR was released in [`v4.229.0`](https://github.com/artsy/cohesion/releases/tag/v4.229.0) :rocket:

PR #557, opened 2025-02-11
feat(EMI-2276): adding estimate_available for when widget fails

xander-pero:

This PR resolves [EMI-2276]

### Description

We've implemented some safeguards for artworks that are marked as being able to have an estimate generated, but actually cannot due to their size/weight/price origin country, disqualifying them from receiving an arta shipping price. This commit adds a field to identify when this occurs so that we can identify when this occurs.

[EMI-2276]: https://artsyproduct.atlassian.net/browse/EMI-2276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -307,23 +307,24 @@ export interface CreateAlertReminderMessageViewed {
 }
 
 /**
- * User sees a shipping estimate
+ * User sees a shipping estimate. estimate_available = false when artwork is disqualified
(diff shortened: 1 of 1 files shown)


erikdstock (approved):

artsyit:
:rocket: PR was released in [`v4.232.0`](https://github.com/artsy/cohesion/releases/tag/v4.232.0) :rocket:

PR #499, opened 2024-05-07
feat: add ClickedBuyNow in events schema

MrSltun:
The type of this PR is: **Enhancement**

This PR involves [EMI-1783]

### Description
This PR adds `ClickedBuyNow` event into the Events schema for Force to match the events we have in Eigen
This is a prerequisite to the work for tracking Parter Offer tracking in Force ([ticket details](https://artsyproduct.atlassian.net/browse/EMI-1783) 🔒)

### PR Checklist (tick all before merging)

- [x] If I've added a new file to the tree I've exported it from the common [`index.ts`](https://github.com/artsy/cohesion/blob/main/src/Schema/Events/index.ts)
- [x] I've added comments with examples for any new interfaces and ensured that they're in the docs
- [x] No platform-specific terminology has been used outside of `click` and `tap` (platform is inferred by the DB storing events)

[EMI-1783]: https://artsyproduct.atlassian.net/browse/EMI-1783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #558, opened 2025-02-11
feat(EMI-2258): add ClickedMakeOffer to events schema for signal tracking

xander-pero:
The type of this PR is: **feature**

This PR resolves [EMI-2258]

### Description

I copied changes from this [PR](https://github.com/artsy/cohesion/pull/499/files) and the current `clickedBuyNow` schema; not sure if we should remove `flow` `signal_lot_watcher_count` and `signal_bid_count`, for future reasons like making offers on a partner offer, and since the latter two still existed on `clickedBuyNow`. 

Also, do these events not need to be declared in the `export type Event = ` beginning on line 263 in [index.ts](https://github.com/artsy/cohesion/blob/bd9c398c83377b74418a7a2b45b94c0448d52458/src/Schema/Events/index.ts#L263) because they exist in DeprecatedSchema?

[EMI-2258]: https://artsyproduct.atlassian.net/browse/EMI-2258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DeprecatedSchema/DeprecatedValues.ts
+++ b/src/DeprecatedSchema/DeprecatedValues.ts
@@ -49,14 +49,7 @@ export enum ActionType {
    * TODO: Check if ‘Tap’ and this can be combined.
    */
   Click = "Click",
-  ClickedBid = 'Clicked "Bid"',
-  /**
-   * A click on 'Buy Now' or 'Make offer' buttons.
-   */
-  ClickedBuyNow = "Clicked buy now",
   ClickedConsign = "Clicked consign",
-  ClickedContactGallery = 'Clicked "Contact Gallery"',
-  ClickedMakeOffer = "Clicked make offer",
 
   /**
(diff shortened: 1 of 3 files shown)


xander-pero:
Added `ClickedBuyNow`, `ClickedBid`, `ClickedContactGallery` and `ClickedMakeOffer` in the exports in `index.ts` and removed those from `DeprecatedValues.ts`

Is there follow-up required in force for `ClickedBuyNow`, `ClickedBid`, and `ClickedMakeOffer` as well, considering that it seems now exists from a different file?

artsyit:
:rocket: PR was released in [`v4.234.0`](https://github.com/artsy/cohesion/releases/tag/v4.234.0) :rocket:

PR #560, opened 2025-02-18
fix: add YourMaxBid owner type

MrSltun:

This PR resolves []

### Description
This PR is a prerequisite to artsy/eigen#11566
This PR adds `yourMaxBid` to cohesion as an owner type, to keep the tracking after the refactoring of the bid flow in artsy/eigen#11522


artsyit:
:rocket: PR was released in [`v4.236.0`](https://github.com/artsy/cohesion/releases/tag/v4.236.0) :rocket:

PR #561, opened 2025-02-19
fix: add ConfirmYourBid owner type

MrSltun:

This PR resolves []

### Description
This PR is a follow-up on #560
This PR adds `confirmYourBid` to cohesion as an owner type, to keep the tracking after the refactoring of the bid flow


xander-pero (approved):
🫡  aye looks good

artsyit:
:rocket: PR was released in [`v4.237.0`](https://github.com/artsy/cohesion/releases/tag/v4.237.0) :rocket:

PR #566, opened 2025-03-12
feat: add tracking data for long press and create alert

dariakoko:

This PR resolves [ONYX-1616]

### Description

Supports PRs https://github.com/artsy/eigen/pull/11653 (add long press tracking, fix Create Alert tracking) and https://github.com/artsy/eigen/pull/11657 (track progressive onboarding tooltips)

[ONYX-1616]: https://artsyproduct.atlassian.net/browse/ONYX-1616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Schema/Events/ImpressionTracking.ts
```diff
@@ -82,6 +82,28 @@ export interface TooltipViewed {
...
+ *  ```
+ *  {
+ *    action: "tooltipViewed",
+ *    context_module: "newWorksForYouRail",
+ *    context_owner_type: "home",
+ *    type: "long-press-artwork-context-menu"
+ *  }
+ * ```
+ */
+export interface TooltipViewedApp {
```
olerichter00, line 100: I don't see the prefix `App` being used. Maybe we can just remove it 🤔 And because this is related to Onboarding, maybe we should add it to the name as well.

olerichter00 (commented):
Could also use the existing `tooltipViewed` event?

src/Schema/Events/ImpressionTracking.ts
dariakoko: `tooltipViewed` interface has some params that we don't eed to mention explicitly on eigen, also, it uses type PageOwnerType which is web specific.
I will rename the `TooltipViewedApp` interface

src/Schema/Events/index.ts
```diff
@@ -1573,6 +1575,10 @@ export enum ActionType {
    * Corresponds to {@link TooltipViewed}
    */
   tooltipViewed = "tooltipViewed",
+  /**
+   * Corresponds to {@link ProgressiveOnboardingTooltipViewed}
+   */
+  progressiveOnboardingTooltipViewed = "progressiveOnboardingTooltipViewed",
```
olerichter00, line 1581: Is this used anywhere? [ProgressiveOnboardingTooltipViewed](https://github.com/artsy/cohesion/pull/566/files#diff-2181c09c8693ccb9a0667bf93e9caf26546fa65a0f99a0002aca0e0dd44e8ad8R100) uses `tooltipViewed`.

dariakoko: omg i got confused with the setup. correct, it should not be there

artsyit:
:rocket: PR was released in [`v4.243.0`](https://github.com/artsy/cohesion/releases/tag/v4.243.0) :rocket:

PR #567, opened 2025-03-12
feat: share lists tracking events

nickskalkin:

### Description

Added a few events to track our experimental "share lists" functionality:

image

We already have a `Share` event which allows us to track when user clicks the "Copy URL" button, but we were missing clicks on the "Share" button itself, and the "Open in the new tab" button.

I've also added an event to track when someone opens shared collections.

### DIFF
--- a/src/Schema/Events/ArtworkLists.ts
+++ b/src/Schema/Events/ArtworkLists.ts
@@ -115,3 +115,23 @@ export interface ViewedArtworkList {
   context_owner_type: OwnerType.saves
   owner_id: string
 }
+
+/**
+ * When a user clicks to view a shared artwork list
+ *
+ * This schema describes events sent to Segment from [[viewedSharedArtworkList]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "viewedSharedArtworkList",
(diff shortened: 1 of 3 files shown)


src/Schema/Events/ArtworkLists.ts
```diff
@@ -115,3 +115,23 @@ export interface ViewedArtworkList {
   context_owner_type: OwnerType.saves
   owner_id: string
 }
+
+/**
+ * When a user clicks to view a shared artwork list
```
dariakoko, line 120: **Question:** what do you think about rephrasing it to match with the action type?
`When a user views a shared artwork list?`

dariakoko (approved):

evaschilken:
Thanks nikita! Do we also track the page view when someone opens a shared list?

src/Schema/Events/ArtworkLists.ts
nickskalkin: Thanks, will do!

artsyit:
:rocket: PR was released in [`v4.241.0`](https://github.com/artsy/cohesion/releases/tag/v4.241.0) :rocket:

PR #571, opened 2025-03-19
feat: add interface for TappedMenuItemGroup event

dblandin:
### Description

To track a user selecting an item within a menu group, for example the navigation menu in the Account screen.

Will eventually replace the `tappeMenuItem` event only used in a couple places.


louislecluse (approved):

artsyit:
:rocket: PR was released in [`v4.245.0`](https://github.com/artsy/cohesion/releases/tag/v4.245.0) :rocket:

PR #572, opened 2025-03-24
fix(batch-import): Use action instead of event

damassi:
The type of this PR is: **Fix**

### Description

We were getting errors over on the client about the analytics schema, and realized we _need_ an action for each event. And also added a missing highlight warning/error event, while re-adding the error one. 

cc @artsy/amber-devs


artsyit:
:rocket: PR was released in [`v4.246.0`](https://github.com/artsy/cohesion/releases/tag/v4.246.0) :rocket:

PR #573, opened 2025-03-24
feat: add express checkout viewed event

rquartararo:

### Description

This adds an `expressCheckoutViewed` event to track when a user views the express checkout in the order flow. 


xander-pero:
Event definition looks good. Triple checking that we're talking about the same thing for firing logic:
this event would fire when a user lands on the `Shipping` page, and Express Checkout button is available, like in screenshot:
Screenshot 2025-03-24 at 3 57 34 PM

xander-pero (approved):

PR #574, opened 2025-03-25
chore(EMI-2366): update express checkout viewed to receive multiple payment methods

rquartararo:

This PR resolves [EMI-2366]

### Description
This updates the express checkout viewed to receive multiple payment methods as when we start adding more express checkout options the user could view multiple checkout options / buttons. 

[EMI-2366]: https://artsyproduct.atlassian.net/browse/EMI-2366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


xander-pero (approved):
Good catch!

xander-pero:
Yep, no problem at all. Can be like that

PR #575, opened 2025-03-27
chore: Rename wallet type field and correct example values

xander-pero:

### Description

Exchange will store the `payment_method` as `credit_card` for Express Checkout orders, but created a new field called `credit_card_wallet_type` which will describe the type of Express Checkout used (applePay, googlePay). This renames the field on the event to match.


artsyit:
:rocket: PR was released in [`v4.249.0`](https://github.com/artsy/cohesion/releases/tag/v4.249.0) :rocket:

PR #579, opened 2025-04-10
feat(DIA-1252): add tappedShare event

iskounen:
This PR adds a `tappedShare` event that can be fired from various surfaces when users share artworks, articles, artists, collections, auctions, etc.


artsyit:
:rocket: PR was released in [`v4.253.0`](https://github.com/artsy/cohesion/releases/tag/v4.253.0) :rocket:

PR #583, opened 2025-04-14
feat: add artworkActions to ContextModule

iskounen:
This PR adds `artworkActions` to the `ContextModule` enum so that Eigen can use the Cohesion schema instead of its own (deprecated) schema. These actions are currently being used by saves, view-in-room, and shares.


artsyit:
:rocket: PR was released in [`v4.255.0`](https://github.com/artsy/cohesion/releases/tag/v4.255.0) :rocket:

PR #586, opened 2025-04-22
feat: Migrate SubmittedOffer and SubmittedOrder

xander-pero:

- `clicked_express_checkout` proof on the `slug` `id` comment [here](https://artsy.looker.com/sql/bxyvktf78mj2b9).
- `analytics` tables only require the `order_id` field on the `submitted_order` and `submitted_offer` tables.
- We maintain the same event name as the deprecated event `submitted_order` and `submitted_offer`, which I left in DeprecatedSchema in case things go wrong. 
- Importantly, implementation will fire one of these events when completing Express Checkout, which will restore CVR to CA on the Session level. 

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -2423,8 +2423,7 @@ export interface ClickedHeroUnitGroup {
  *  {
  *    action: "clickedExpressCheckout",
  *    context_page_owner_type: "ordersShipping",
- *    context_page_owner_slug: "radna-segal-pearl",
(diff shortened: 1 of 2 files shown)


xander-pero:
@rquartararo I'm not sure how you'll hook on the `order_id`, but going off what was seen on the `clicked_express_checkout` event,  looks like it corresponds to wherever `context_page_owner_id` points on that event.

artsyit:
:rocket: PR was released in [`v4.257.0`](https://github.com/artsy/cohesion/releases/tag/v4.257.0) :rocket:

PR #595, opened 2025-05-19
feat(settingsV2): Adds new events for PartnerLocation edit and new

jpotts244:
The type of this PR is: Feature

This PR resolves @leodmz's tracking audit for the CMS settings V2 migration project [here](https://www.notion.so/artsy/Setting-2-tracking-1f8cab0764a080dc9feff40f7f729485) 

### Description

Adds two new events to be triggered by the new CMS V2 experience
- Add new location event
- Edit existing location even

cc @artsy/amber-devs 


src/Schema/CMS/Events/SettingsFlow.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @example
+ * ```
+ * {
+ *   action: "edited location",
+ *   context_module: "Settings",
+ *   before_address_type: "Business",
+ *   after_address_type: "Business",
+ *   partner_location_id: "some-partner-location-id",
+ *   user_id: "some-user-id",
```
jpotts244, line 35: Just a heads up here @leodmz , the `after_address_type` and `before_address_type` could def be the same value on this event (the user could have updated other fields like `address`, `city` etc etc)


Is this enough data on your end?

leodmz: yes, that is probably the case already

leodmz:
I'm confused because I thought those events were already being fired but I cannot see them in cohesion and we're creating new ones. 
What events were we using before ?

leodmz:
> Volt also has this Legacy tracking service which has defined [these events before](https://github.com/artsy/volt/blob/main/app/javascript/apps/analytics/utils/LegacyTracker.tsx#L72-L74) and called in our legacy location form [here](https://github.com/artsy/volt/blob/75f7b401b2d2d9e438311b1962d0feff5aaa25a8/app/javascript/legacy/components/LocationForm/LocationForm.tsx#L100-L125) which hooks up to the legacy Volt ruby side of things for onboarding and settings locations ([here](https://github.com/artsy/volt/blob/main/app/views/locations/_form.html.haml#L4-L10))
> 
> I figured we wanted to make these more Cohesion based/modern! Let me know, if you have concerns with this being added to cohesion! I think we will have to to make this accessible in the Modern Settings JS

ah! thanks for the clarification, I didn't know about the volt legacy tracking stuff. 
yes I'm all for having them added to cohesion

leodmz (approved):

artsyit:
:rocket: PR was released in [`v4.264.1`](https://github.com/artsy/cohesion/releases/tag/v4.264.1) :rocket:

PR #596, opened 2025-05-29
feat: Add Order Details tracking

xander-pero:

This PR resolves [EMI-2494] and [EMI-2509]

### Description

Added the new TappedBuyerProtection event as described [here](https://docs.google.com/spreadsheets/d/1t-6TEsFCXI00lBAAgBL8wSMeAigO7fORKPmswHoCF5g/edit?gid=0#gid=0). 

Also added a new OwnerType for the OrdersDetail page, which is a new pagetype we are supporting for app and web in [release](https://artsy.slack.com/archives/C02JHHHKP5K/p1748541224172779).

### DIFF
--- a/src/Schema/Events/Tap.ts
+++ b/src/Schema/Events/Tap.ts
@@ -1430,3 +1430,29 @@ export interface TappedMenuItemGroup {
   position: number
   subject: string
 }
+
+/**
+ * A user taps buyer protection on order details page
+ *
+ * This schema describes events sent to Segment from [[tappedBuyerProtection]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "tappedBuyerProtection",
+ *    context_module: "OrdersDetail",
+ *    context_screen_owner_type: "orders-detail",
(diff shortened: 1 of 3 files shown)


src/Schema/Events/Tap.ts
```diff
@@ -1430,3 +1430,29 @@ export interface TappedMenuItemGroup {
...
+ *    action: "tappedBuyerProtection",
+ *    context_module: "OrdersDetail",
+ *    context_screen_owner_type: "orders-detail",
+ *    context_screen_owner_id: "57e60c68-a198-431e-8a02-6ecb01e3a99b",
+ *    destination_screen_owner_type: "articles",
+ *    destination_screen_owner_slug: "360048946973-How-does-Artsy-protect-me"
+ *  }
+ * ```
+ */
+export interface TappedBuyerProtection {
```
oxaudo, line 1451: we currently already have `ClickedBuyerProtection ` - this one is to just distinguish web/app?

Also minor but seems like they are calling it 'Buyer Guarantee' across the board now. Wonder if we should rename this event as well?

oxaudo (approved):
Looks good to me. 

One minor question about naming, but otherwise good to go for me.

PR #604, opened 2025-06-13
fix(bulk-edits): Adds more consistent typing for number typed values

jpotts244:

### Description

Small improvement, adjust how we handle "number" values in the CMS Bulk edit events

Adds more consistent typing for number typed values, we have a mix of string and number usage 

cc @artsy/amber-devs


artsyit:
:rocket: PR was released in [`v4.272.1`](https://github.com/artsy/cohesion/releases/tag/v4.272.1) :rocket:

PR #605, opened 2025-06-13
fix: Adjust bulkEditFailed type to more match error structure

jpotts244:

### Description

Small adjustment to the typing to better match how we are formatting errors from websockets over in volt

Ex:
```
[ 'availability: must be "for sale", "sold" or "not for sale" for works available for purchase' ]
```

cc @artsy/amber-devs 


artsyit:
:rocket: PR was released in [`v4.272.2`](https://github.com/artsy/cohesion/releases/tag/v4.272.2) :rocket:

PR #606, opened 2025-06-16
feat: Adds artwork_ids metadata to processing and shortlist bulk edit events

jpotts244:

This PR resolves Feedback from QA with @leodmz 

### Description

Adds `artwork_ids` metadata to processing and shortlist bulk edit events

-  [x] I've added comments with examples for any new interfaces and ensured that they're in the docs


jpotts244:
going in! feel free to add feedback

PR #607, opened 2025-06-18
feat: Additional metadata for processingCompleted and click availability bulk edit flow

jpotts244:

This PR resolves [AMBER-1705](https://artsyproduct.atlassian.net/browse/AMBER-1705)

### Description

Adding additional metadata to our bulk edits
- `click change availability`
- `processing completed`

events

cc @artsy/amber-devs 

[AMBER-1705]: https://artsyproduct.atlassian.net/browse/AMBER-1705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/CMS/Events/BulkEditFlow.ts
+++ b/src/Schema/CMS/Events/BulkEditFlow.ts
@@ -175,7 +175,7 @@ export interface CmsBulkEditResolvedAllConflictsShown {
 }
 
 /**
- * Bluk edit flow processing started
+ * Bulk edit flow processing started
  *
  * Example:
  * {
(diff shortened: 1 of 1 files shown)


src/Schema/CMS/Events/BulkEditFlow.ts
```diff
@@ -195,14 +197,16 @@ export interface CmsBulkEditProcessingStarted {
  * {
  *   action: "processingCompleted",
  *   context_module: "Artworks - bulk edit",
- *   value: 24  // total number of artworks successfully processed
+ *   label: "change availability",
+ *   value: "on hold", // e.g. "on hold", "available", "not for sale"
```
jpotts244, line 201: `label` and `value` will change based on the three options we have (so far)

Label can be:
```
- change availability
- change location
- change medium
```

and then value will be any of the selected options for the field they are editing
```
Ex:
- An availability: "on hold"
- An address: "123 Happy St, NY NY 10009"
- A medium: "Painting"
```

PR #611, opened 2025-06-27
fix: add destination type to TappedBuyerProtection

MrSltun:
The type of this PR is: **Fix**

This PR resolves part of [EMI-2509]

### Description
This PR adds `destination_screen_owner_type` to the `TappedBuyerProtection` event type 


xander-pero:
Sorry I missed this! Thanks 👍

PR #613, opened 2025-07-02
feat: Orders Detail native app tracking

xander-pero:

### Description

This PR adds two new events: `TappedImportFees` and `OrderDetailsViewed`. The former is extending `ClickedImportFees` to app, and the second is a new impression event to track the displayed content to the user. We could not add this to the pageview event.

Some notes:
- Because `OrderDetailsViewed` will fire on web and app, I use `OwnerType` and therefore do not specify the screen/page within the field or OwnerType.
- We believe `message_type`, which corresponds to the displayed content, is sufficient to describe the message displayed to the user. I thought about recording the delivery (shipping/pickup), arta vs. flat/free, and payment method here, but we have the `order_id` and can join that info easily. The message may change, but those order components should stay static.
- @MrSltun ClickedImportFees has a `flow` field, so I've added that to the Tapped version as well. Comparing this to Tapped/Clicked`BuyerProtection`, that does not have a `flow`. For simplicity I'm mirroring the Click event so that the web/app versions have the same fields. `ClickedVisitHelpCenter` and `clickedAskSpecialist` also have `flow`, and so does `TappedAskSpecialist`.

### DIFF
--- a/src/Schema/CMS/Events/BatchImportFlow.ts
+++ b/src/Schema/CMS/Events/BatchImportFlow.ts
@@ -233,9 +233,11 @@ export interface CmsBatchImportTableContentSummary {
   context_module: CmsContextModule.batchImportFlow
   context_page_owner_type: CmsOwnerType.batchImport
   context_page_owner_id: string
-  values: Record<string, { optional: boolean; valid: number; missing: number; invalid: number }>
+  values: Record<
+    string,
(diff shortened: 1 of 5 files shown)


MrSltun:
Looks great! I'll merge it now so I could add the tracking events on Eigen now

PR #616, opened 2025-07-14
feat(onyx-1828): add artworkRecommendations as PageOwnerType

nickskalkin:

This PR resolves [ONYX-1828]

### Description

We are adding artwork recommendations rail to the web. `OwnerType.artworkRecommendations` is already a part of a `ScreenOwnerType` union, now making it a part of `PageOwnerType`.


artsyit:
:rocket: PR was released in [`v4.283.0`](https://github.com/artsy/cohesion/releases/tag/v4.283.0) :rocket:

PR #625, opened 2025-07-28
feat: Add ClickedGraphCTA

xander-pero:

This PR resolves [CO-]

### Description

We are adding new CTAs to the bottom of a few graphs on the analytics page which will re-direct the partner to artworks, conversations, orders. This PR adds tracking for these using a single event, `clickedGraphCTA`, which will use `ContextModule` to discern which graph the user is interacting with. `Destination_page_owner_type` is also there.


leamotta (approved):

artsyit:
:rocket: PR was released in [`v4.291.0`](https://github.com/artsy/cohesion/releases/tag/v4.291.0) :rocket:

PR #628, opened 2025-08-04
feat: add ViewedToast interface

dariakoko:

This PR resolves [ONYX-1812]

### Description

Add toast tracking to track a toast introduced in this ticket: https://artsyproduct.atlassian.net/browse/ONYX-1812
- add ViewedToast interface

[ONYX-1812]: https://artsyproduct.atlassian.net/browse/ONYX-1812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -113,15 +113,15 @@ export interface ProgressiveOnboardingTooltipViewed {
  *  ```
  *  {
  *    action: "viewedToast",
- *    context_module: "artworkRecommendations",
- *    context_screen_owner_type: "artworkGrid",
+ *    context_module: "artworkGrid",
+ *    context_screen_owner_type: "artworkRecommendations",
  *    subject: "Set an artwork budget in your profile at any time."
  *  }
  * ```
  */
 export interface ViewedToast {
   action: ActionType.viewedToast
-  context_module: ContextModule
+  context_module?: ContextModule
   context_screen_owner_type: OwnerType
   subject: string
 }
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -105,6 +105,28 @@ export interface ProgressiveOnboardingTooltipViewed {
 }
 
 /**
+ * A user viewed a toast. `subject` will describe what toast it is.
+ *
+ * This schema describes events sent to Segment from [[viewedToast]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "viewedToast",
+ *    context_module: "artworkRecommendations",
+ *    context_screen_owner_type: "artworkGrid",
+ *    subject: "Set an artwork budget in your profile at any time."
+ *  }
+ * ```
+ */
+export interface ViewedToast {
+  action: ActionType.viewedToast
+  context_module: ContextModule
+  context_screen_owner_type?: OwnerType
+  subject: string
+}
+
+/**
  * A user sees a an error message
  *
  * This schema describes events sent to Segment from [[ErrorMessageViewed]].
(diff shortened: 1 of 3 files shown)


src/Schema/Events/ImpressionTracking.ts
```diff
@@ -104,6 +104,28 @@ export interface ProgressiveOnboardingTooltipViewed {
...
+ *    context_module: "artworkRecommendations",
+ *    context_screen_owner_type: "artworkGrid",
+ *    subject: "Set an artwork budget in your profile at any time."
+ *  }
+ * ```
+ */
+export interface ViewedToast {
+  action: ActionType.viewedToast
+  context_module: ContextModule
+  context_screen_owner_type?: OwnerType
```
dblandin, line 125: I can imagine there would always be a screen associated with this view event. Maybe we make this required and update the example to use the home view?

I don't know if we would often have an additional context module as these toasts usually float above the screen you're on. Is that attribute necessary in this case?

dariakoko: for the context, the usage will look like this:
```
...
      context_screen_owner_type: OwnerType.artworkRecommendations,
      context_module: ContextModule.artworkGrid,
...
```

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -104,6 +104,28 @@ export interface ProgressiveOnboardingTooltipViewed {
...
+ *    context_screen_owner_type: "artworkGrid",
+ *    subject: "Set an artwork budget in your profile at any time."
+ *  }
+ * ```
+ */
+export interface ViewedToast {
+  action: ActionType.viewedToast
+  context_module: ContextModule
+  context_screen_owner_type: OwnerType
+  subject: string
```
MounirDhahri, line 126: I wonder if `subject` is needed, I would have assumed we would be saving an `id` that doesn't change regardless of the copy

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -104,6 +104,28 @@ export interface ProgressiveOnboardingTooltipViewed {
...
+ * A user viewed a toast. `subject` will describe what toast it is.
+ *
+ * This schema describes events sent to Segment from [[viewedToast]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "viewedToast",
+ *    context_module: "artworkRecommendations",
+ *    context_screen_owner_type: "artworkGrid",
```
dblandin, line 117: These example values look swapped, right? The screen owner type is `artworkRecommendation` while the intended context module is `artworkGrid`?

src/Schema/Events/ImpressionTracking.ts
dblandin: Gotcha. I think the example values are swapped at the moment. I see `artworkGrid` used for `context_screen_owner_type.

And I suppose `artworkGrid` makes sense as that's where the trigger logic is found – we fire the toast in response to scroll depth on the grid.

src/Schema/Events/ImpressionTracking.ts
dblandin: I think subject is fine here – we use this pattern for other buttons / toasts where it's useful to have the copy context but we don't really need a consistent identifier across copy changes. I don't think we often use the `subject` for querying. If we ever _do_ need consistent query support – perhaps something like `analyticsID` would be an appropriate addition to these schema events – but I haven't heard a need/desire for that yet.

dblandin (approved):
LGTM (but please update the example before merge!) – I misunderstood the use of contextModule, which looks fine now, but the values are stills swapped in the example.

artsyit:
:rocket: PR was released in [`v4.293.0`](https://github.com/artsy/cohesion/releases/tag/v4.293.0) :rocket:

PR #629, opened 2025-08-04
chore: Make a generic click event for batch edits, mark others for deprecation

jpotts244:
The type of this PR is: **CHORE**

### Description

I'm about to add another click event to the batch edits tracking schema: `Add to show`
Before adding another custom event, how about we make the existing click events under Batch Edits more generic so we dont have to update cohesion moving forward

WDYT?

I'll be leaving the specific events in the schema for now and just mark them as deprecated while we update the client to use the new more general event

cc @artsy/amber-devs


artsyit:
:rocket: PR was released in [`v4.295.0`](https://github.com/artsy/cohesion/releases/tag/v4.295.0) :rocket:

PR #634, opened 2025-09-02
chore: Update tapped and clicked complete your profile

xander-pero:

### Description

[slack thread](https://artsy.slack.com/archives/C02JHHHKP5K/p1756327122075039)with context.

We recently added a "Complete your profile" link on web for the order details page. We'd like to track interaction with it, reusing the logic from the existing event. 

Upon further investigation, we realized that we don't have any data for these events, potentially because they were not being imported in `index.ts` here. I've added those calls, and also deleted some of the subfields since we didn't see any existing force/eigen implementation that would cause an issue.

Note: This does not solve the fact that `tappedCompleteYourProfile` _is not_ firing when I tap on the "Complete my profile" button on the App on the profile screen. I simply updated it to be the new schema to match `clickedCompleteYourProfile`, and provided an example of what I'd expect it to look like. 

Screenshot 2025-09-02 at 6 03 12 PM

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -2372,19 +2372,26 @@ export interface ClickedVisitHelpCenter {
 }
 
 /**
- * A user clicks on the complete your profile prompt within the activity pannel
+ * A user clicks on the complete your profile on the order details page (no activity panel)
  *
  * This schema describes events sent to Segment from [[clickedCompleteYourProfile]]
  *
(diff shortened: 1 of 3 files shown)


leamotta (approved):

artsyit:
:rocket: PR was released in [`v4.299.0`](https://github.com/artsy/cohesion/releases/tag/v4.299.0) :rocket:

PR #691, opened 2026-04-09
feat: more single page checkout tracking

xander-pero:

### Description

This does the following:
- adds `default_address_id` and `default_address_country` fields to the `savedAddressViewed`, so that we can know which was the default address saved, and what quotes would've generated upon landing.
- adds `shippingQuoteViewed` which will track impressions on the shipping quotes. We will definitely do this for arta, but I kept the naming open in case we expand providers or want to include flat/free

    - **Decided not to implement tracking when a user completes saving an address. If we fix our firing of the `savedAddressViewed` step, this event would be synonymous with `clickedOrderProgression` when the user navigates from Fulfillment step to the next.**

cc @artsy/emerald-devs 

### DIFF
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -424,6 +424,7 @@ export interface ExpressCheckoutViewed {
  *    context_page_owner_id: "6164889300d643000db86504",
  *    flow: "Buy now" | "Make offer" | "Partner offer"
  *    address_ids: ["1234abcd5678efgh9012ijkl"]
+ *    default_address_id: "1234abcd5678efgh9012ijkl"
  *  }
  * ```
  */
@@ -433,6 +434,7 @@ export interface SavedAddressViewed {
   context_page_owner_id: string
   flow: string
   address_ids: string[]
+  default_address_id: string
 }
 
 /**
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -488,6 +488,32 @@ export interface NewPaymentMethodViewed {
 }
 
 /**
+ * A user views a retrieved shipping quote in checkout
+ *
+ * This schema describes events sent to Segment from [[shippingQuoteViewed]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "shippingQuoteViewed",
+ *    context_page_owner_type: "orders-checkout",
+ *    context_page_owner_id: "6164889300d643000db86504",
+ *    flow: "Buy now" | "Make offer" | "Partner offer"
+ *    shipping_quotes: ["Standard", "Express"]
+ *    shipping_quote_ids: ["abcd1234efgh5678ijkl9012", "mnop3456qrst7890uvwx1234"]
+ *  }
+ * ```
+ */
+export interface ShippingQuoteViewed {
+  action: ActionType.shippingQuoteViewed
+  context_page_owner_type: PageOwnerType
+  context_page_owner_id: string
+  flow: string
+  shipping_quotes: string[]
+  shipping_quote_ids: string[]
+}
+
+/**
  * A user views a step in checkout
  *
  * This schema describes events sent to Segment from [[OrderProgressionViewed]]
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -501,8 +501,10 @@ export interface NewPaymentMethodViewed {
  *    context_page_owner_type: "orders-checkout",
  *    context_page_owner_id: "6164889300d643000db86504",
  *    flow: "Buy now" | "Make offer" | "Partner offer"
- *    shipping_quotes: ["Standard", "Express"]
- *    shipping_quote_ids: ["abcd1234efgh5678ijkl9012", "mnop3456qrst7890uvwx1234"]
+ *    shipping_quotes: [
+ *      { id: "ABC", type: "standard", price_minor: 123, price_currency: "USD", timeline: "Est. delivery: 3–5 days after shipping" },
+ *      { id: "XYZ", type: "express", price_minor: 456, price_currency: "USD", timeline: "Est. delivery: 2 days after shipping" },
+ *    ]
  *  }
  * ```
  */
@@ -511,8 +513,13 @@ export interface ShippingQuoteViewed {
   context_page_owner_type: PageOwnerType
   context_page_owner_id: string
   flow: "Buy now" | "Make offer" | "Partner offer"
-  shipping_quotes: string[]
-  shipping_quote_ids: string[]
+  shipping_quotes: Array<{
+    id: string
+    type: string
+    price_minor: number
+    price_currency: string
+    timeline: string
+  }>
 }
 
 /**
--- a/src/Schema/Events/ImpressionTracking.ts
+++ b/src/Schema/Events/ImpressionTracking.ts
@@ -425,6 +425,7 @@ export interface ExpressCheckoutViewed {
  *    flow: "Buy now" | "Make offer" | "Partner offer"
  *    address_ids: ["1234abcd5678efgh9012ijkl"]
  *    default_address_id: "1234abcd5678efgh9012ijkl"
+ *    default_address_country: "US"
  *  }
  * ```
  */
@@ -435,6 +436,7 @@ export interface SavedAddressViewed {
   flow: string
   address_ids: string[]
   default_address_id: string
+  default_address_country: string
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/Schema/Events/ImpressionTracking.ts
```diff
@@ -485,6 +487,32 @@ export interface NewPaymentMethodViewed {
...
+ *    shipping_quotes: ["Standard", "Express"]
+ *    shipping_quote_ids: ["abcd1234efgh5678ijkl9012", "mnop3456qrst7890uvwx1234"]
+ *  }
+ * ```
+ */
+export interface ShippingQuoteViewed {
+  action: ActionType.shippingQuoteViewed
+  context_page_owner_type: PageOwnerType
+  context_page_owner_id: string
+  flow: string
```
MounirDhahri: minor: to make sure we don't accidentally send strings formatted differently and end up making your life harder 😄 
```suggestion
  flow: "Buy now" | "Make offer" | "Partner offer"
```

xander-pero: I was under the impression it's best practice to keep these open, rather than have hardcoded values (for the potential future where we introduce new options).

In your opinion, is hardcoding these worthwhile compared against the pain of potentially needing to keep it free later on? 

This specific one is tiny, but thinking about my future cohesion PRs.

MounirDhahri: > I was under the impression it's best practice to keep these open, rather than have hardcoded values (for the potential future where we introduce new options).

I wasn't aware of that but in this case, I would actually push towards having the flow string typed for a couple of reasons:
- leverage typescript type safety
  - no other string value slips (example `BuyNow`, `Buy now`)
  - type completion inside IDE
 - do not allow sending values that do not match to anything (example "")

> In your opinion, is hardcoding these worthwhile compared against the pain of potentially needing to keep it free later on?

In this case, I think it would be more painful to keep this as a free string. These strings do not change so often, so hardcoding them could actually protect us from sending the wrong strings in client apps. And if at some point we want to add a new flow, we can simply add it here and in client apps then bump cohesion version. 

I would personally only use `string` if the set of values is genuinely open-ended or determined at runtime (e.g., user-generated content, values from an external API we don't control, id...).

MounirDhahri (approved):

MounirDhahri:
Thank you ❤️

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -485,6 +489,32 @@ export interface NewPaymentMethodViewed {
...
+ * This schema describes events sent to Segment from [[shippingQuoteViewed]]
+ *
+ *  @example
+ *  ```
+ *  {
+ *    action: "shippingQuoteViewed",
+ *    context_page_owner_type: "orders-checkout",
+ *    context_page_owner_id: "6164889300d643000db86504",
+ *    flow: "Buy now" | "Make offer" | "Partner offer"
+ *    shipping_quotes: ["Standard", "Express"]
```
starsirius, line 504: "Standard" or "Express" could still be ambiguous because domestic and international shipping share the same label. You could probably join the order and line item to differentiate, but I wonder if we can support that directly in the event, such as including the timeframe description (which is what the buyers see and an important driver), e.g. "Est. delivery 3-5 days after shipping".

Image

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -485,6 +489,32 @@ export interface NewPaymentMethodViewed {
...
+ *    shipping_quote_ids: ["abcd1234efgh5678ijkl9012", "mnop3456qrst7890uvwx1234"]
```
starsirius, line 505: #question: `shipping_quotes` and `shipping_quote_ids` are independent arrays which might need to be kept in the same order if/when zipping. Curious if we have a pattern for more complex objects like `ShippingQuote` which includes an id, label, and description. And then we can send an array of `ShippingQuote` in the event and hopefully that's easier to act upon.

xander-pero: Let me browse. I introduced this set up with `savedPaymentMethodViewed` event [here](https://github.com/artsy/cohesion/blob/ea8a8bd16a2c3d19e514975025ad420239678177/src/Schema/Events/ImpressionTracking.ts#L460), but I agree this isn't the best way to do it

xander-pero: I wonder if our ELT set up would even allow that right now. Would it then be stored as a double-layered JSON in a single column, that I then need to extract? Definitely not a problem, but more headache than single JSON

xander-pero: Ok, with 10 mins of digging I think what we do with `commercialFilterParamsChanged` ([example](https://github.com/artsy/cohesion/blob/ea8a8bd16a2c3d19e514975025ad420239678177/src/Schema/Events/FilterAndSort.ts#L23)) or any of these filtering events to be the best. It _looks_ like it's just a string extract of a formatted JSON. 

Something like:
`{"shippingQuoteIDs":["ABC","XYZ","123"],"shippingQuotes":["standard","express","premium"],"shippingQuotePricesMinor":[123,456,789],"shippingQuotePriceCurrencies":["USD","USD","USD"]}`

Neatified:
`
{
"shippingQuoteIDs":["ABC","XYZ","123"],
"shippingQuotes":["standard","express","premium"],
"shippingQuotePricesMinor":[123,456,789],
"shippingQuotePriceCurrencies":["USD","USD","USD"]
}
`

@starsirius Thoughts on this?

Screenshot 2026-04-14 at 3 55 41 PM
Screenshot 2026-04-14 at 3 58 10 PM

xander-pero:
@starsirius @erikdstock I'd like to merge this in and for us to start tracking these changes asap. Lmk if I should change the set up for the shipping quote schema outlined above, or if we're good to proceed with the current approach.

src/Schema/Events/ImpressionTracking.ts
starsirius: @xander-pero Not exactly. The issue from the above implementation that uses multiple arrays (e.g. `"shippingQuoteIDs":["ABC","XYZ","123"], "shippingQuotes":["standard","express","premium"]`), and to access a quote, we will find `shippingQuoteIDs[0]`, `shippingQuotes[0]`, `shippingQuotePricesMinor[0]`, by the same index. The downside is that it requires the quote order in those arrays should be exactly the same, and it's easier for them to get out of sync.

Usually a preferred data structure would be an array of objects, so something like this:

```json
{
  "shippingQuotes": [
    {
      "ID": "ABC",
      "type": "standard",
      "priceMinor": 123
    },
    {
      "ID": "XYZ",
      "type": "express",
      "priceMinor": 456
    },
    {
      "ID": "123",
      "type": "premium",
      "priceMinor": 789
    }
  ]
}
```

BUT, I wonder if we already have patterns to work with this shape of data on the analytics side. Certainly don't want to introduce something that's disruptive.

src/Schema/Events/ImpressionTracking.ts
starsirius: @xander-pero How do you feel about including the shipping timeframe copy in the event?

xander-pero: Yeah let's do it, including the raw copy is probably best if we ever want to mess with it, and we can extract the min and max using substrings or something

src/Schema/Events/ImpressionTracking.ts
xander-pero: Hmm, this would be completely new for `cohesion` but it's not an issue afaik? I believe it would get cast to a `SUPER` [type](https://docs.aws.amazon.com/redshift/latest/dg/r_SUPER_type.html) in redshift, which we then can extract payload information ([degas example](https://github.com/artsy/degas/blob/76f72d3caaa6232e5cc6c330e15c234f17f546c8/dbt/models/analytics/auctions/sales.sql#L45)) at any nested level. 

I'm on board with trying this

xander-pero:
@starsirius just reworked to use an `Array`. Also, I did a quick search in the repo and found this [example](https://github.com/artsy/cohesion/blob/ea8a8bd16a2c3d19e514975025ad420239678177/src/Schema/CMS/Events/CompletenessScoreFlow.ts#L56) that already uses an `Array`, so I'm assuming this'll work without a hitch too.

src/Schema/Events/ImpressionTracking.ts
```diff
@@ -485,6 +489,40 @@ export interface NewPaymentMethodViewed {
...
+  context_page_owner_id: string
+  flow: "Buy now" | "Make offer" | "Partner offer"
+  shipping_quotes: Array<{
+    id: string
+    type: string
+    subtype: string
+    price_minor: number
+    price_currency: string
+    timeline: string
+  }>
```
starsirius, line 523: Could be formalized to a type like `ShippingQuoteType`. Not blocking.

starsirius (approved):
Excited for more impression tracking!

artsyit:
:rocket: PR was released in [`v4.353.0`](https://github.com/artsy/cohesion/releases/tag/v4.353.0) :rocket:

PR #694, opened 2026-05-18
feat: add one-tap auth trigger #trivial

brainbicycle:

This PR resolves [DI-291](https://www.notion.so/artsy/6-Force-One-Tap-Basic-Analytics-361cab0764a0804c9a28d989f4c950fe?v=310cab0764a08151b558000cc3bc6840&source=copy_link)

### Description

- adds one-tap as an AuthTrigger to differentiate from regular google sign in 
- ~~adds context_page_path in auth events to capture page auth started on~~

https://artsy.slack.com/archives/C05EEBNEF71/p1778881199224469?thread_ts=1778874112.044779&cid=C05EEBNEF71

- [x] No platform-specific terminology has been used outside of `click` and `tap` (platform is inferred by the DB storing events) - maybe?

### DIFF
--- a/src/Schema/Events/Authentication.ts
+++ b/src/Schema/Events/Authentication.ts
@@ -48,7 +48,6 @@ export interface AuthImpression {
  *    action: "createdAccount",
  *    auth_redirect: "https://artsy.net/artist/andy-warhol",
  *    context_module: "popUpModal",
- *    context_page_path: "/artist/andy-warhol",
  *    intent: "viewArtist",
  *    modal_copy: "Sign up to follow artists",
  *    onboarding: true,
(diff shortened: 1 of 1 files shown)


xander-pero:
AFAIK `context_page_path` is actually something we already get on all our events based on the default web set up. I'm not sure how this works on the backend, but I see this firing within all the payloads for `track` events and is also present on the [table](https://artsy.looker.com/sql/7tr7zf8vrsv4bc). So just the AuthTrigger change is necessary!
Screenshot 2026-05-18 at 2 40 59 PM

Screenshot 2026-05-18 at 2 36 20 PM

artsyit:
:rocket: PR was released in [`v4.355.0`](https://github.com/artsy/cohesion/releases/tag/v4.355.0) :rocket:

PR #695, opened 2026-05-21
chore: use new auth method field rather than expanding triggers

brainbicycle:

This PR resolves [CO-]

### Description

Switch trigger out for a new method field as it does not really fit conceptually with other triggers. 
Analytics pr that is still not merged is only thing passing the modified trigger so should be okay to remove. 


xander-pero (approved):
lgtm thanks!

artsyit:
:rocket: PR was released in [`v4.356.0`](https://github.com/artsy/cohesion/releases/tag/v4.356.0) :rocket:

PR #701, opened 2026-06-10
feat: add country field to auto complete

xander-pero:

### Description

Upon QAing async, I realized we don't actually capture the country within the payloads! This makes sense because we launched to US only, but we now must capture the Country we're trying to complete for in order to look at anything and everything!

@artsy/emerald-devs what makes sense here for a string value? I assumed we would just use the Country Code (and that's my preference),  but I am also completely fine with us using the string "United States" or whatever is visible to the user. Those would always be unique and 1:1 right?

### DIFF
--- a/src/Schema/Events/Search.ts
+++ b/src/Schema/Events/Search.ts
@@ -290,6 +290,7 @@ export interface SelectedSearchSuggestionQuickNavigationItem {
  *    context_module:"OrdersShipping",
  *    context_owner_type: "orders-shipping",
  *    context_owner_id: "57e60c68-a198-431e-8a02-6ecb01e3a99b",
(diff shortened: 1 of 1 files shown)


starsirius (approved):
🎣

artsyit:
:rocket: PR was released in [`v4.358.0`](https://github.com/artsy/cohesion/releases/tag/v4.358.0) :rocket:

src/Schema/Events/Search.ts
```diff
@@ -326,6 +329,7 @@ export interface SelectedItemFromAddressAutoCompletion {
   context_module: ContextModule
   context_owner_type: PageOwnerType
   context_owner_id?: string
+  country: string
```
MounirDhahri, line 332: maybe we can be less defensive here
```suggestion
  country?: string
```

PR #703, opened 2026-06-16
chore: add partner offer convo event

araujobarret:

### Description

Add a partner offer conversation viewed event. We aim to send these events once the partner offer CTA is visible for partners and collectors.

cc @artsy/topaz-devs 


src/Schema/Events/PartnerOffer.ts
```diff
@@ -0,0 +1,27 @@
...
+ *  @example
+ *  ```
+ *  {
+ *    action: "partnerOfferInConversationViewed",
+ *    context_owner_type: "conversations"
+ *    context_owner_id: "55002921"
+ *  }
+ * ```
+ */
+export interface PartnerOfferInConversationViewed {
```
leodmz: This should be added to ImpressionTracking.ts instead of a new folder

src/Schema/index.ts
```diff
@@ -19,6 +19,7 @@ export * from "./Events/ImpressionTracking"
 export * from "./Events/Inquiries"
 export * from "./Events/MyCollection"
 export * from "./Events/MyCollectionInsights"
+export * from "./Events/PartnerOffer"
```
leodmz: This is not needed if you add the event to ImpressionTracking.ts

src/Schema/Events/index.ts
```diff
@@ -1325,6 +1325,10 @@ export enum ActionType {
    * Corresponds to {@link OrderProgressionViewed}
    */
   orderProgressionViewed = "orderProgressionViewed",
+  /**
+   * Corresponds to {@link PartnerOfferInConversationViewed}
+   */
+  partnerOfferInConversationViewed = "partnerOfferInConversationViewed",
```
leodmz, line 1333: There are actually 2 other places in this doc where a new event should be added. 
L212 - to the list of events contained within impressiontracking
L619 - to the full list of events being exported

leodmz (changes requested):
thank you for putting this together. 
I would prefer for the event to be in ImpressionTracking to ensure we follow cohesion taxonomy 
+ some required changes in index.ts

src/Schema/Events/PartnerOffer.ts
araujobarret: I went there first and changed the idea after seeing this comment at the beginning of ImpressionTracking.ts:
```
/**
 * Schemas describing rail_viewed events
 * @packageDocumentation
 */
```

leodmz: right, that rail_viewed event part is not accurate considering the range of events in the folder. 
This is where we've defaulted to define all user side `_viewed` events

araujobarret:
Thanks for the review @leodmz , I've moved it to the ImpressionTracking.ts and covered the other requests in the last commit 👍

leodmz (approved):

jpotts244 (commented):
Nice!!

artsyit:
:rocket: PR was released in [`v4.360.0`](https://github.com/artsy/cohesion/releases/tag/v4.360.0) :rocket:

PR #704, opened 2026-06-18
chore: make address autocomplete Country optional

xander-pero:

### Description

This was blocking a Topaz release because we have not implemented the `country` call in Force yet. [Thread](https://artsy.slack.com/archives/C02BC3HEJ/p1781772094691399).


araujobarret (approved):

artsyit:
:rocket: PR was released in [`v4.361.0`](https://github.com/artsy/cohesion/releases/tag/v4.361.0) :rocket:

PR #709, opened 2026-06-24
chore: Add tracking schema for the invetory table

MrSltun:

This PR resolves this [Notion Card]

### Description
This PR adds a new schema for the inventory table in Art OS

cc @artsy/amber-devs 

_Assisted by: Claude Sonnet 4.6_

[Notion Card]: https://app.notion.com/p/artsy/Implement-tracking-for-Inventory-Table-380cab0764a08114b496ff322362d0f5

<details>
  <summary>📦 Published PR as canary version: <code>4.365.0--canary.709.14936.0</code></summary>
  <br />

  ```bash
  npm install @artsy/cohesion@4.365.0--canary.709.14936.0
  # or 
  yarn add @artsy/cohesion@4.365.0--canary.709.14936.0
  ```
</details>


src/Schema/os/Events/InventoryTable.ts
```diff
@@ -0,0 +1,334 @@
...
+  /** Omit for privateNotes (free text — don't log contents) */
+  old_value?: string
+  /** Omit for privateNotes (free text — don't log contents) */
+  new_value?: string
+  /** Medium only: true = free text entered, false = predefined mediumType selected */
+  is_free_text?: boolean
+  /** Artist and location only: the action taken on the autocomplete */
+  value?: "add" | "matched" | "removed"
+  /** Artist and location only: the resolved Artsy entity ID */
+  matched_id?: string
```
olerichter00, line 132: Should we add here as well whether sync is on or not when the update happens?

src/Schema/os/Events/InventoryTable.ts
```diff
@@ -0,0 +1,334 @@
...
+  context_module: OsContextModule.actionsDropdown
+  context_page_owner_type: OsOwnerType
+  value:
+    | "Convert to Edition Set"
+    | "Convert to Unique"
+    | "Delete"
+    | "Distribute to Artsy"
+    | "Edit Edition Set"
+  /** Which trigger opened the dropdown */
+  label: "action button" | "context menu" | "row action"
```
olerichter00: Which actions dropdown is this about, and is this event still accurate given that we have the "Add to," the "Open in Studio," the "..." and the row action dropdown? 🤔

Image

src/Schema/os/Events/InventoryTable.ts
MrSltun: I believe that will be covered in another task, [here 🔐](https://app.notion.com/p/artsy/Analytics-Tracking-for-Distribution-Sync-Mapping-37bcab0764a081e1ab13e8d28c433b62)

src/Schema/os/Events/InventoryTable.ts
MrSltun: Good point, maybe we can be more granular here, I'll update that

src/Schema/os/Events/InventoryTable.ts
```diff
@@ -0,0 +1,334 @@
...
+ * ```
+ * {
+ *   action: "clickedArtworkRow",
+ *   context_module: "artworkTable",
+ *   context_page_owner_type: "inventory",
+ *   artwork_id: "abc123"
+ * }
+ * ```
+ */
+export interface OsClickedArtworkRow {
```
leodmz, line 26: Are we adding OS because there's already an equivalent in CMS? 
There's no OS prefix in the [notion doc](https://www.notion.so/artsy/Analytics-Tracking-for-Inventory-Table-37bcab0764a081b6ab64c77debfd74df?source=copy_link)

src/Schema/os/Values/OsOwnerType.ts
```diff
@@ -4,5 +4,6 @@
  * @packageDocumentation
  */
 export enum OsOwnerType {
+  collection = "collection",
```
leodmz: Is collection the "list" page? 
I see "lists" in the url - we should probably be consistent with the url here

src/Schema/os/Events/InventoryTable.ts
MrSltun: I just added it to follow the CMS convention that we have for type names (for example, [BatchImport](https://github.com/artsy/cohesion/blob/main/src/Schema/CMS/Events/BatchImportFlow.ts) and [Onboarding](https://github.com/artsy/cohesion/blob/main/src/Schema/CMS/Events/OnboardingFlow.ts)), so when we use it in the front-end, it's clearly labelled in the name but I don't have a strong opinion

leodmz: I wouldn't follow the CMS conventions. 
IMO we can rely on the `context_page_path` to flag if it's an OS or CMS event so I don't think it needs to be in the name

MrSltun: Oh this is for us engineers when we use the type in the front-end 😅 The event action will be `clickedArtworkRow` (check the example in the comment above), but the interface name (`OsClickedArtworkRow`) is for us when we use the event detail when tracking

leodmz: right, no concerns from my side then

leodmz (approved):

src/Schema/os/Values/OsOwnerType.ts
MrSltun: cc @mzikherman (since you recently added `collection`)

olerichter00 (approved):
🌟

artsyit:
:rocket: PR was released in [`v4.365.0`](https://github.com/artsy/cohesion/releases/tag/v4.365.0) :rocket:

PR #721, opened 2026-06-29
feat: counter offer tracking

xander-pero:

This PR resolves [AS-5655]

### Description

Adds tracking for **five** new events we introduce: `clickedCounterOfferOption`, `SubmittedCounterOffer` (was a DeprecatedSchema event), `AcceptedOffer`, `DeclinedOffer`, and `ToggledOfferHistory`. We have a deprecated `SubmittedCounterOffer` -> `submitted_counter_offer` event already, so I follow that convention of **two words.**

I also bit the bullet and added an `AcceptedOffer` and `DeclinedOffer` event, **which we previously didn't fire events for.** This gives me better modeling ability in Degas, rather than hacking together something with the most recent `ClickedOfferOption` event and a `ordersReview ClickedOrderProgression` event. 

_I am consciously naming it AcceptedOffer and DeclinedOffer, not AcceptedCounterOffer and DeclinedCounterOffer, to potentially keep it reusable for C2C, which I would want to reuse all existing events for, rather than introducing a new set of C2C specific events. The 5 events: `SubmitOrder`, `SubmitOffer`, `SubmitCounterOffer`, `AcceptedOffer`, `DeclinedOffer` would be the full extent of events needed for our traditional e-com implementation._

[AS-5655]: https://artsyproduct.atlassian.net/browse/AS-5655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Schema/Events/Click.ts
+++ b/src/Schema/Events/Click.ts
@@ -2653,7 +2653,7 @@ export interface SubmittedOffer {
 }
 
 /**
- * A user submits a counter offer in response to a gallery counter offer. Does not fire for accept/reject actions.
+ * A user submits a counter offer in response to a gallery counter offer.
  *
  * This schema describes events sent to Segment from [[SubmittedCounterOffer]]
  *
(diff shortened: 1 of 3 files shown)


src/Schema/Events/Click.ts
```diff
@@ -811,6 +811,34 @@ export interface ClickedOfferOption {
...
+ *  }
+ * ```
+ */
+export interface ClickedCounterOfferOption {
+  action: ActionType.clickedCounterOfferOption
+  context_module: ContextModule
+  context_page_owner_type: PageOwnerType
+  context_page_owner_id: string
+  option: "accept" | "counter" | "decline"
+  amount?: number
```
xander-pero, line 838: Only applicable to accept/counter options. Null for decline

src/Schema/Events/Click.ts
```diff
@@ -2624,6 +2652,28 @@ export interface SubmittedOffer {
   credit_card_wallet_type?: string
 }
 
+/**
+ * A user submits a counter offer in response to a gallery counter offer. Does not fire for accept/reject actions.
```
xander-pero: Fires onSubmit (aka when the user exits the counter offer flow, in the same we do for one page checkout) rather than onClick. Only fires when the user is countering the gallery offer. Does not fire if they accept or reject.

artsyit:
:rocket: PR was released in [`v4.380.0`](https://github.com/artsy/cohesion/releases/tag/v4.380.0) :rocket:

PR #728, opened 2026-07-03
feat: add edition_set_id to EditedInventoryField event

chr-tatu:
Adds an optional \`edition_set_id\` field to the \`EditedInventoryField\` ArtOS event so edition-set edits can report which edition set the edited value belongs to.

### Description
- Adds \`edition_set_id?: string\` to the \`EditedInventoryField\` interface (\`src/Schema/os/Events/InventoryTable.ts\`), documented as the edition set the edited value belongs to when the edit targets an edition set row.
- Updates the TypeDoc \`@example\` block to include \`edition_set_id\`.
- Adds a test case covering the event with \`edition_set_id\` set.

The field is optional since only edition-set row edits carry it; regular artwork edits omit it.

Ref: [Notion — Add edition_set_id EditedInventoryField and send it from edition events](https://app.notion.com/p/392cab0764a0807fbc55d6954b11d10d)

🤖 Generated with [Claude Code](https://claude.com/claude-code)
cc @artsy/amber-devs

### DIFF
--- a/src/Schema/os/Events/InventoryTable.ts
+++ b/src/Schema/os/Events/InventoryTable.ts
@@ -371,6 +371,7 @@ export interface OsClickedActionsDropdown {
  *   context_page_owner_type: "inventory",
  *   destination: ["artsy"],
  *   artwork_id: "abc123",
+ *   edition_set_id: "xyz789",
  *   field: "availability",
  *   did_push_to_cms: false
  * }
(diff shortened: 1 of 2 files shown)


MrSltun (approved):

artsyit:
:rocket: PR was released in [`v4.384.0`](https://github.com/artsy/cohesion/releases/tag/v4.384.0) :rocket:

PR #729, opened 2026-07-07
fix: allow CompletedArtworkDistribution to fire on any owner type

chr-tatu:
## Summary
- `CompletedArtworkDistribution` was hardcoded to `context_page_owner_type: OsOwnerType.inventory`, so it could only be fired from the inventory surface.
- Widen the field to `OsOwnerType` (matching the sibling `DistributedArtworks` event) so the event can also fire from `collection`, consistent with the ArtOS convention of differentiating surfaces via owner type rather than a module field.
- Added a test asserting the event serializes correctly with `context_page_owner_type: "collection"`.

🤖  Created with Claude Code
@artsy/amber-devs

### DIFF
--- a/src/Schema/os/Events/Submit.ts
+++ b/src/Schema/os/Events/Submit.ts
@@ -105,7 +105,7 @@ export interface DistributedArtworks {
 export interface CompletedArtworkDistribution {
   action: OsActionType.completedArtworkDistribution
   context_module: OsContextModule.artworkTable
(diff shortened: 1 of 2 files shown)


MrSltun (approved):

artsyit:
:rocket: PR was released in [`v4.385.0`](https://github.com/artsy/cohesion/releases/tag/v4.385.0) :rocket:

PR #731, opened 2026-07-07
chore: add oneTap account creation Intent

xander-pero:

### Description

Need: our Account Creation Intent [graph](https://artsy.looker.com/explore/analytics/session_experiments?qid=rygT6fDT3K5NYiwGcc7Ovo&origin_space=1276&toggle=dat,vis) is having a strong `signup` bucket, coming from one tap. We want this graph to split out One Tap into its own group, because this is our main way of actually grouping user intents. 

Screenshot 2026-07-07 at 10 50 01 AM

**Solution 1, discussed on call:** Add a new `AuthIntent` for `oneTap`, which will now log this as its own intent rather than `signup`. (Implemented in this PR)

**Solution 2, PREFERRED after thinking more about it:** Manually account for this in `analytics.users` [model](https://github.com/artsy/degas/blob/dfbbae5b6f804caa5c307f7c490ef8f50fc095de/dbt/models/analytics/core_tables/users.sql#L319) using the `coalesced_account_creation_intent` field, which is where this graph is coming from. I will now order as: `one_tap, account_creation_intent, sign_up_intent`
- Why: the raw event table will still maintain the `signup` intent, which really is the user's intent, but this graph will display `one-tap` as its own separate category. 
<br>
Screenshot 2026-07-07 at 10 54 11 AM
<br>
<br>
@brainbicycle let me know what you think about solution 2. It solves the business need while keeping our `intent` field the same—which feels preferable because `one-tap` would have been recorded twice otherwise.

### DIFF
--- a/src/Schema/Values/Intent.ts
+++ b/src/Schema/Values/Intent.ts
@@ -19,6 +19,7 @@ export enum Intent {
   seePriceAuctionRecords = "seePriceAuctionRecords",
   seeRealizedPriceAuctionRecords = "seeRealizedPriceAuctionRecords",
   makeOffer = "makeOffer",
+  oneTap = "oneTap",
   registerToBid = "registerToBid",
   requestConditionReport = "requestConditionReport",
   saveArtwork = "saveArtwork",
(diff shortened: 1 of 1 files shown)


brainbicycle (approved):
solution 2 sounds reasonable to me! let me know if you do want the analytics change on my end though 🙏

xander-pero:
Will close this PR in favor of Solution 2; so no backend code changes needed I believe!

xander-pero:
@brainbicycle can you shoot me an approval on that [degas PR](https://github.com/artsy/degas/pull/1108) I tagged you on?

PR #738, opened 2026-08-17
feat: add private vr events

dariakoko:

This PR resolves [CO-]

### Description
Adding data for ArtOS private viewing rooms tracking 

cc @artsy/amber-devs


artsyit:
:rocket: PR was released in [`v4.391.0`](https://github.com/artsy/cohesion/releases/tag/v4.391.0) :rocket:

PR #742, opened 2026-08-28
feat: add suggestedFilters context module

egdbear:

### Description
[Force's search dropdown](https://github.com/artsy/force/pull/17606) adds a suggested filters row: a parsed medium, price and rarity suggestion linking to a filtered /collect page. This module attributes its impressions and clicks, which would otherwise be reported as topTab and be indistinguishable from entity results in the same dropdown.


artsyit:
:rocket: PR was released in [`v4.394.0`](https://github.com/artsy/cohesion/releases/tag/v4.394.0) :rocket:

PR #749, opened 2026-09-18
feat: add Art Assistant events

nickskalkin:

### Description

Adds the analytics schema for*Art Assistant (the conversational search flow in the app). The app ships it behind `AREnableArtAssistant` and currently answers with an artwork rail. Artists, shows, viewing rooms and galleries follow, so the turn events describe the entities generically (`item_type` / `item_ids`) rather than naming artworks.

**New values**

- `OwnerType.artAssistant` (screen owner type; app-only, so it is not added to `PageOwnerType`)
- `ContextModule.artAssistant`, `ContextModule.artAssistantResults`, `ContextModule.artAssistantSuggestions`

**New events**

| Event | Fires when |
| --- | --- |
| `tappedArtAssistant` | A user taps an entry point into the flow. `type: "icon" \| "button"` names the affordance, `context_module` names where it lives — mirrors `tappedSearchByImage` for Artsy Lens. |
| `tappedArtAssistantSuggestion` | A user taps a suggested prompt on the empty state (fills the composer). Carries the suggestion copy and its position. |
| `tappedArtAssistantNewChat` | A user confirms discarding a chat, with `message_count` of what was discarded — also a signal the conversation did not deliver. |
| `sentArtAssistantMessage` | A user sends a message. Carries `conversation_id`, `message_id`, `message_index`, `character_count` and whether the text was typed or came from a suggestion. |
| `receivedArtAssistantResponse` | The agent finishes a turn. Carries `duration_ms`, `tool_call_count`, `stop_reason` and the entities shown (`item_type`, `item_ids`, `item_count`). `item_type` is free text, named after the matching `OwnerType`, so the app can start showing a new entity type without a schema change. |
| `artAssistantTurnFailed` | A turn ends without an answer, with `outcome` separating a broken stream, a stalled stream, a turn that closed empty, an agent that gave up, and a missing session. `error_status` carries the HTTP status (401, 429, …) when there was one. |

**Message content**

`sentArtAssistantMessage.message` and `receivedArtAssistantResponse.response` are optional and hold user/agent words, so the app only populates them while a separate feature flag is on — the rest of both events is always sent. `response` holds the final answer only, without the intermediate activity updates ("Thinking…", "Searching artworks…"). If we are told not to collect the content, the flag goes off and every metric above keeps working.

`message_id` joins the turn together: `receivedArtAssistantResponse.prompt_message_id` and `artAssistantTurnFailed.prompt_message_id` both point back to it.

**Item taps** reuse the existing `tapped*Group` events (`tappedArtworkGroup` today, `tappedArtistGroup` / `tappedShowGroup` / … as entity types land) with `context_module: "artAssistantResults"` and `context_screen_owner_type: "artAssistant"`, so assistant-driven taps show up in existing per-entity dashboards. No new event for those.


artsyit:
:rocket: PR was released in [`v4.401.0`](https://github.com/artsy/cohesion/releases/tag/v4.401.0) :rocket:

PR #750, opened 2026-09-18
feat: add artist instagram tracking events

rquartararo:

### Description

Tracking for the artist page "Social" rail and the CMS handle entry that feeds it. Two new
events, plus two new `ContextModule` values.

**Web**:

- `clickedSocialPost` — a user clicks an Instagram post in the "Social" rail on the artist
  overview page (`ArtistSocialRail`). `destination_path` is optional because `permalink` is
  nullable and the tile renders `href={permalink ?? undefined}`, so a linkless tile is possible.

**CMS** — added to the existing `ArtistFlow`:

- `enteredArtistSocialHandle` — a partner saves an artist's handle in the "Socials" section of
  the artist overview page. `outcome` distinguishes a first entry from an edit.

New values: `ContextModule.socialRail` and `CmsContextModule.artistOverviewSocials`. No
existing enum values changed, so nothing downstream moves.

**Edits to three original events**

- *Rail impression* — no new event. `useRailImpressionTracking({ contextModule:
  ContextModule.socialRail })` fires the existing shared `railViewed`, with dwell detection and
  once-per-page-view dedupe already handled. This only needed the enum value.
- *Header handle link* — already shipping. `ArtistHeaderSocialLink` fires `clickedHeader` with
  `subject: "social"`.

Names are service-agnostic with a `service` field (`"instagram"` today) so a second network
becomes a union member rather than a new event and a new Redshift table. It is a constant value
for now, since `instagramMedia` is Instagram-only.

**Drive-by doc fixes** (both shipped docs, both wrong):

- `RailViewed`'s `@example` claimed `action: "rail_viewed"`; the enum is `railViewed`.
- `ClickedHeader`'s `@example` omitted `subject: "social"`, which force already sends.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


starsirius:
💪

xander-pero:
AFAIK we haven't implemented rail_viewed much on web, I think only on NWFY (maybe?) to cover impressions for recommendations freshness. 

I'm fine for us to add that for the purposes of the launch, but worth keeping in mind long term behavior of that so that our expected behavior on rails across web is consistent

xander-pero (changes requested):
Sorry I'm on Mobile so having a hard time navigating. Is there a reason we introduced a new Social.ts instead of sticking within Click.ts? That can handle the new Click event you outlined

rquartararo:
@xander-pero good catch that was added when I thought we needed multiple social events but then I ended up reusing other events. I will fold it into Click.ts.

xander-pero (approved):

artsyit:
:rocket: PR was released in [`v4.402.0`](https://github.com/artsy/cohesion/releases/tag/v4.402.0) :rocket:

